PR #1, opened 2021-11-22
Create 0001-library-adr.rst

kenclary:
Draft of ADR for repo.


connorhaugh (approved):
Looks good to me, seems to capture what we want with this without being too restrictive.

PR #2, opened 2021-11-24
feat: setup repo as NPM package with edX standards.

kenclary:


openedx.yaml
```diff
@@ -0,0 +1,8 @@
+# openedx.yaml
+
+---
+owner: edx/fedx-team
```
davidjoy, line 4: May want to change this to T&L, though I'm not sure if it is used/matters.

davidjoy (approved):
LGTM 🚢

src/setupTest.js
```diff
@@ -0,0 +1,29 @@
...
+process.env.MARKETING_SITE_BASE_URL = 'http://localhost:18000';
+process.env.ORDER_HISTORY_URL = 'localhost:1996/orders';
+process.env.REFRESH_ACCESS_TOKEN_ENDPOINT = 'http://localhost:18000/login_refresh';
+process.env.SEGMENT_KEY = 'segment_whoa';
+process.env.SITE_NAME = 'edX';
+process.env.USER_INFO_COOKIE_NAME = 'edx-user-info';
+process.env.LOGO_URL = 'https://edx-cdn.org/v3/default/logo.svg';
+process.env.LOGO_TRADEMARK_URL = 'https://edx-cdn.org/v3/default/logo-trademark.svg';
+process.env.LOGO_WHITE_URL = 'https://edx-cdn.org/v3/default/logo-white.svg';
+process.env.FAVICON_URL = 'https://edx-cdn.org/v3/default/favicon.ico';
```
doctoryes, line 29: NIT: I compared this list to the one here:
https://github.com/edx/frontend-platform/blob/master/.env.development
and there appear to be some missing values. Not sure if it's important at all.

doctoryes (approved):
Just a question/nit.

src/setupTest.js
kenclary: these files were based on the configuration of frontend-component-header, which is a recent example of a frontend library.

PR #3, opened 2021-12-03
fix: update node to 16, using .nvmrc file, to support later versions of semantic-release.

kenclary:


.github/workflows/commitlint.yml
```diff
@@ -7,4 +7,4 @@ on:
 
 jobs:
   commitlint:
-    uses: edx/.github/.github/workflows/commitlint.yml@main
+    uses: edx/.github/.github/workflows/commitlint.yml@master
```
doctoryes, line 10: Q: So this line isn't referring to the `commitlint.yml` file in this repo? This repo has a `main` branch but not `master`.

doctoryes (approved):
Just one question.

.github/workflows/commitlint.yml
kenclary: correct. the older repo still uses master

PR #9, opened 2022-01-06
feat: Text Editor and V2 Editor Framework

connorhaugh:


connorhaugh:
@kenclary @doctoryes This is ready for your review at your earliest connivence, then I will get a React expert's review.

src/editors/EditorPage.jsx
```diff
@@ -0,0 +1,72 @@
...
+        onClose={() => {}}
+        hasCloseButton={false}
+        variant="dark"
+      >
+        <div className="d-flex flex-column vh-100">
+          <EditorHeader title={blockType} />
+          {selectEditor(blockType)}
+          <EditorFooter />
+        </div>
+
```
kenclary, line 57: maybe remove extraneous whitespace

kenclary (approved):
Looks good. (minor comment included)

src/editors/EditorFooter.jsx
```diff
@@ -0,0 +1,77 @@
...
+    unitUrlLoading,
+    unitUrl,
+    setSaveUnderway,
+    saveUnderway,
+    saveResponse,
+    studioEndpointUrl,
+    editorRef,
+  } = useContext(EditorPageContext);
+
+  const onSaveClicked = () => {
```
muselesscreator, line 26: These are easier to test if you define them outside the component in the same file :-)

src/editors/EditorHeader.test.jsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import EditorHeader from './EditorHeader';
+import EditorPageContext from './EditorPageContext';
+import { ActionStates } from './data/constants';
+
+delete window.location;
+window.location = {
```
davidjoy, line 9: You may want to be careful with doing this work at the top of the file like this, as it'll get evaluated when the file is loaded, not necessarily when the test is run.  You may want to instead put this into a `beforeEach` and then clean up after yourself afterward by putting the `assign` function back.

src/editors/EditorPageProvider.jsx
```diff
@@ -0,0 +1,88 @@
...
+        courseId,
+        studioEndpointUrl,
+        blockContent,
+        { setInProgress: setSaveUnderway, setResponse: setSaveResponse },
+      );
+    }
+  }, [saveUnderway]);
+
+  return (
+    <EditorPageContext.Provider
```
davidjoy, line 53: So I'd be careful here.  This provider will cause everything subscribing to it in its `children` to re-render on every single re-render here.  It's creating a new Object on every re-render, which trips all the `useContext` hooks below it in the tree, even if the data they depend on in the context hasn't changed.  Worst case scenario, this can cause a massive increase in re-rendering and slow your page down.

Context is best for data that changes infrequently, and when you're passing a lot of data into it like this, it may be best to memoize creation of the `value` so that you're not creating a new context value object each time this provider re-renders, if that makes sense.

davidjoy: I may not have explained well here - I'm happy to do a better job synchronously if it's helpful.

src/editors/EditorHeader.test.jsx
connorhaugh: I'll plan to clean these up! Thanks

src/editors/EditorPageProvider.jsx
connorhaugh: No that makes sense-- Its not really feasible to say don't use `useEffect` hooks in your editors, so I'll plan to use the `useMemo` hook around the value, and do some testing with logging as to how often things rerender. It seems to be effective at solving that problem, as we can make the dependancies of the memo hook just the loading states

connorhaugh: Alright so using memoization, I have `EditorFooter` (the consumer component which uses an effect hook) rendering on only 50% of renders from editorprovider, and the effect hooks are called much less often. Thanks!

src/editors/data/api.test.js
```diff
@@ -0,0 +1,66 @@
...
+  setLoading: jest.fn(),
+};
+const saveFunctionsSave = {
+  setResponse: jest.fn(),
+  setInProgress: jest.fn(),
+};
+const blockId = 'coursev1:2uX@4345432';
+const studioEndpointUrl = 'hortus.coa';
+
+test('fetchBlockById 404', () => {
```
doctoryes, line 19: `404` -> `401`

src/editors/data/api.test.js
```diff
@@ -0,0 +1,66 @@
...
+  expect(saveFunctionsGet.setError).toHaveBeenCalled();
+});
+test('fetchUnitById 404', () => {
+  get.mockRejectedValue({ response: { status: 404 } });
+  fetchUnitById(saveFunctionsGet, blockId, studioEndpointUrl);
+  expect(saveFunctionsGet.setLoading).toHaveBeenCalled();
+  expect(saveFunctionsGet.setError).toHaveBeenCalled();
+});
+test('saveBlock 408', () => {
+  post.mockRejectedValue({ response: { status: 404 } });
```
doctoryes, line 56: `404` -> `408`

src/editors/data/constants.js
```diff
@@ -0,0 +1,31 @@
...
+    case 'html':
+      /**/
+      return {
+        id: blockId,
+        category: blockType,
+        has_changes: true,
+        data: content,
+        couseKey: courseId,
+      };
+    default: return {};
```
doctoryes, line 29: Nit: Perhaps if a block is not handled in one of the `case` statements above, this code should complain more noisily?

doctoryes (commented):
First review complete.

src/editors/EditorFooter.test.jsx
```diff
@@ -0,0 +1,107 @@
...
+  };
+  render(
+    <EditorPageContext.Provider value={context}>
+      <EditorFooter />
+    </EditorPageContext.Provider>,
+  );
+
+  expect(screen.getByText('Add To Course')).toBeTruthy();
+  userEvent.click(screen.getByText('Add To Course'));
+  expect(saveBlock).not.toHaveBeenCalled(); // not called by footer, called in provider
```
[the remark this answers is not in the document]
connorhaugh: I will look into that

src/editors/EditorPageProvider.jsx
```diff
@@ -0,0 +1,93 @@
...
+    saveUnderway,
+    studioEndpointUrl,
+    blockId,
+    courseId,
+    blockType,
+  }), [blockLoading, unitUrlLoading, saveUnderway]);
+
+  useEffect(() => {
+    if (unitUrlLoading === ActionStates.NOT_BEGUN) {
+      fetchUnitById({ // pass as object for testing purposes.
```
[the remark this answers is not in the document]
connorhaugh: yes, testing purposes means "unit testing" purposes

src/editors/data/constants.js
connorhaugh: Yeah I agree!

src/editors/EditorPage.jsx
```diff
@@ -0,0 +1,72 @@
...
+        title={blockType}
+        isOpen
+        size="fullscreen"
+        onClose={() => {}}
+        hasCloseButton={false}
+        variant="dark"
+      >
+        <div className="d-flex flex-column vh-100">
+          <EditorHeader title={blockType} />
+          {selectEditor(blockType)}
```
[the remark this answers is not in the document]
connorhaugh: Yeah it will! curly braces can get you a lot in React

src/editors/EditorPageProvider.jsx
```diff
@@ -0,0 +1,93 @@
...
+  const [blockError, setBlockError] = useState(null);
+  const [blockLoading, setBlockLoading] = useState(ActionStates.NOT_BEGUN);
+  const [unitUrl, setUnitUrlValue] = useState(null);
+  const [unitUrlError, setUnitUrlError] = useState(null);
+  const [unitUrlLoading, setUnitUrlLoading] = useState(ActionStates.NOT_BEGUN);
+  const [blockContent, setBlockContent] = useState(null); // This is the updated content to be saved via api call
+  const [saveResponse, setSaveResponse] = useState(null);
+  const [saveUnderway, setSaveUnderway] = useState(ActionStates.NOT_BEGUN);
+
+  const value = useMemo(() => ({
```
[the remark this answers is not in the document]
connorhaugh: Will do

src/editors/EditorPageProvider.jsx
```diff
@@ -0,0 +1,93 @@
...
+    saveResponse,
+    setSaveUnderway,
+    saveUnderway,
+    studioEndpointUrl,
+    blockId,
+    courseId,
+    blockType,
+  }), [blockLoading, unitUrlLoading, saveUnderway]);
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
connorhaugh: So we want to run "useeffect" as little as possible. We only want it to run:
1. when we initialize, to start the fetching of data
2. When the save process has been begun. 
The second argument to useEffect is an array of dependancies, which trigger a call to useEffect async when they are updated. So, we set our dependency to be `saveUnderway` as it has the following lifecycle:
a. On page load, it is set to value `not_begun`, triggering the fetches. (solving for 1.)
b. On the click of the save button (in footer) saveUnderway = `in_progress` which satifes 2.

If we added more dependancies, it would run every time they updated, causing more re-renders of the all the children of the provider. Does that make sense?

src/editors/EditorFooter.test.jsx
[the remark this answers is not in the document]
connorhaugh: pushed up all other changes, refactoring to make this possible. I thin ben W's comment is the solution here

src/editors/EditorPageProvider.jsx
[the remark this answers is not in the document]
connorhaugh: Should I add a comment detailing this choice?

doctoryes (approved):
👍🏼

src/editors/EditorFooter.jsx
```diff
@@ -50,16 +50,16 @@ export default function EditorFooter() {
...
       />
       </Toast>
       )}
       <ModalDialog.Footer>
         <ActionRow>
           <ActionRow.Spacer />
-          <Button variant="tertiary" onClick={onCancelClicked}>Cancel</Button>
-          <Button onClick={onSaveClicked} aria-label="Save">
+          <Button aria-label="Discard Changes and Return to Learning Context" variant="tertiary" onClick={onCancelClicked}>Cancel</Button>
+          <Button aria-label="Save Changes and Return to Learning Context" onClick={onSaveClicked}>
```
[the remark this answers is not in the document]
connorhaugh: Great question! I did a little with chrome dev tools just now, and it seems passable. We'll get review from Jeff before full rollout though.

PR #10, opened 2022-01-27
fix: dont encode with uri twice

connorhaugh:
Just as the title says-- block key comes in encoded, don't encode it again. Will require a version bump in course authoring upon merge/release

### DIFF
--- a/src/editors/data/api.js
+++ b/src/editors/data/api.js
@@ -36,7 +36,7 @@ export async function fetchUnitById(updateContext, blockId, studioEndpointUrl) {
 
 export async function saveBlock(blockId, blockType, courseId, studioEndpointUrl, content, updateContext) {
   const normalizedContent = normalizeContent(blockType, content, blockId, courseId);
-  const url = `${studioEndpointUrl}/xblock/${encodeURI(blockId)}`;
+  const url = `${studioEndpointUrl}/xblock/${blockId}`;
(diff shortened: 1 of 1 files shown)


src/editors/data/api.js
```diff
@@ -36,7 +36,7 @@ export async function fetchUnitById(updateContext, blockId, studioEndpointUrl) {
 
 export async function saveBlock(blockId, blockType, courseId, studioEndpointUrl, content, updateContext) {
   const normalizedContent = normalizeContent(blockType, content, blockId, courseId);
-  const url = `${studioEndpointUrl}/xblock/${encodeURI(blockId)}`;
+  const url = `${studioEndpointUrl}/xblock/${blockId}`;
```
doctoryes, line 39: Did you try the changed POST locally? Did the save still work? If so, any ideas why it works both ways locally - but not on stage-edx?

connorhaugh: I did try the post locally, after re-installing my local version of the package, as well as with the github version of the package (doesn't have the change) Both worked exactly the same. It frankly doesn't make any sense.

doctoryes: Not understanding the behavior worries me. But this PR is for non-enabled functionality and might shed more light on what's happening. So 🚢 🇮🇹

doctoryes (approved):

PR #11, opened 2022-01-28
feat: adding some tinymce plugins to TextEditor.

kenclary:


doctoryes:
@kenclary Could you build a sandbox where we could try the text editor plugins and see how they're laid out?
@connorhaugh Would it be possible to view the new text editor on a sandbox? What configuration would need to be changed?

connorhaugh:
@doctoryes Viewing a local Version of the npm package in the sandbox would be interesting. Here's how you can do it you would first need to make a branch in course authoring mfe, which contains the module-config.js file, with the path to the /src/ folder which is a sibling folder of edx-platform and all the other idas. You then need to make a sandbox, get it to succeed, then go into the src folder and clone in this branch in there.

What might be easier is to merge this, release the npm package, then create a branch on the course-authoring-mfe which uses that package which we then can look at. For review, I'd love some screenshots or maybe a screen recoding @kenclary

connorhaugh:
Closing this in favor of: https://github.com/edx/frontend-lib-content-components/pull/12

PR #13, opened 2022-02-11
Feat: edit xblock titles

rayzhou-bit:
This features allows users to edit and save xblock titles for the new editor.

There are 3 states:

**loading**
image

**loaded but not editing**
image

**loaded and editing**
image

### DIFF
--- a/src/editors/EditorFooter.jsx
+++ b/src/editors/EditorFooter.jsx
@@ -65,7 +65,7 @@ export default function EditorFooter() {
               : (
                 <FormattedMessage
                   id="authoring.editorfooter.savebutton.label"
-                  defaultMessage="Add To Course"
+                  defaultMessage="Save"
                   description="Label for Save button"
                 />
               )}
--- a/src/editors/EditorHeader.jsx
+++ b/src/editors/EditorHeader.jsx
@@ -1,15 +1,17 @@
-import {
-  ActionRow, IconButton, Icon, ModalDialog, Form,
-} from '@edx/paragon';
 import React, { useContext, useState, useEffect } from 'react';
+import PropTypes from 'prop-types';
+import { FormattedMessage } from '@edx/frontend-platform/i18n';
+import {
+  ActionRow, Form, IconButton, Icon, ModalDialog,
+} from '@edx/paragon';
 import { Edit, Close } from '@edx/paragon/icons';
 import EditorPageContext from './EditorPageContext';
-import { ActionStates } from './data/constants';
+import { ActionStates, mapBlockTypeToName } from './data/constants';
 
 /* The Header component renders based on three states:
     loading -> loaded but not editing -> editing. */
 
-const EditorHeader = () => {
+const EditorHeader = ({ blockType }) => {
   const {
     titleRef, editorRef, setBlockTitle,
     blockValue, blockLoading,
@@ -17,29 +19,28 @@ const EditorHeader = () => {
   } = useContext(EditorPageContext);
 
   const [editing, setEditing] = useState(false);
-
-  const [title, setTitle] = useState('');
+  const [title, setTitle] = useState(mapBlockTypeToName(blockType));
   useEffect(() => {
     if (blockLoading === ActionStates.FINISHED) {
-      setTitle(blockValue ? blockValue.data.display_name : '');
+      setTitle(blockValue ? blockValue.data.display_name : mapBlockTypeToName(blockType));
     }
   }, [blockLoading]);
 
-  const handleChange = (e) => {
-    setTitle(e.target.value);
+  const updateTitle = () => {
+    /* This is called when the input field loses focus.
+       It will update the title we are saving and set editing to false. */
+    setBlockTitle(title);
+    setEditing(false);
   };
 
   const handleKeyDown = (e) => {
-    if (e.key === 'Enter') { setEditing(false); }
+    if (e.key === 'Enter') {
+      setEditing(false);
+    }
     if (e.key === 'Tab' && editorRef) {
       e.preventDefault();
       editorRef.current.focus();
     }
-  };
-
-  const handleOnBlur = () => {
-    setBlockTitle(title);
-    setEditing(false);
   };
 
   const onCancelClicked = () => {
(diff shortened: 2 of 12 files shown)


src/editors/EditorHeader.jsx
```diff
@@ -1,27 +1,82 @@
...
+  useEffect(() => {
+    if (blockLoading === ActionStates.FINISHED) {
+      setTitle(blockValue ? blockValue.data.display_name : '');
+      setEditing(true);
+    }
+  }, [blockLoading]);
+
+  const handleChange = (e) => {
+    setTitle(e.target.value);
+  };
```
rayzhou-bit, line 31: Updates state for title when user edits title.

rayzhou-bit, line 31: Updates state for title when user edits title.

src/editors/EditorHeader.jsx
```diff
@@ -1,27 +1,82 @@
...
+
+  const handleKeyDown = (e) => {
+    if (e.key === 'Enter') setEditing(false);
+    if (e.key === 'Tab' && editorRef) {
+      e.preventDefault();
+      editorRef.current.focus();
+    }
+  };
```
rayzhou-bit, line 39: When user is editing title, they will exit the editing process if they press 'Enter'. If they press tab, the cursor should move on to the editor.

src/editors/EditorHeader.jsx
```diff
@@ -48,52 +50,67 @@ const EditorHeader = () => {
...
+            onChange={handleChange}
+            onKeyDown={handleKeyDown}
+            onBlur={handleOnBlur}
+            autoFocus
+            trailingElement={<Icon src={Edit} />}
+          />
+        </Form.Group>
+      );
+    }
+  }
```
rayzhou-bit, line 88: Implemented this to fix lint error.

src/editors/EditorFooter.jsx
```diff
@@ -25,9 +25,7 @@ export default function EditorFooter() {
```
rayzhou-bit, line 25: This commit holds minor changes as well as updates for test and lint errors.

src/editors/EditorFooter.jsx
```diff
@@ -65,7 +65,7 @@ export default function EditorFooter() {
               : (
                 <FormattedMessage
                   id="authoring.editorfooter.savebutton.label"
-                  defaultMessage="Add To Course"
+                  defaultMessage="Save"
```
rayzhou-bit, line 68: Blocks could be saved to libraries instead of just the course so this button is now changed to "Save".

src/editors/EditorHeader.jsx
```diff
@@ -1,44 +1,112 @@
...
+    setTitle(e.target.value);
+  };
+
+  const handleKeyDown = (e) => {
+    if (e.key === 'Enter') { setEditing(false); }
+    if (e.key === 'Tab' && editorRef) {
+      e.preventDefault();
+      editorRef.current.focus();
+    }
+  };
```
rayzhou-bit, line 38: Handles various key inputs from user. Pressing 'tab' while editing the title will bring the cursor to the editor.

src/editors/EditorHeader.jsx
```diff
@@ -1,44 +1,112 @@
...
+            onChange={handleChange}
+            onKeyDown={handleKeyDown}
+            onBlur={handleOnBlur}
+            autoFocus
+            trailingElement={<Icon src={Edit} />}
+          />
+        </Form.Group>
+      );
+    }
+  }
```
rayzhou-bit, line 88: The key logic for what is displayed in the title.
When the block is loading, this area should only show 'Loading...'.
When the user is not editing, this area will display the text with an edit button following it.
When the user is editing, this area will display a text input.

src/editors/EditorHeader.jsx
```diff
@@ -1,44 +1,112 @@
...
+    blockValue, blockLoading,
+    unitUrl, unitUrlLoading, studioEndpointUrl,
+  } = useContext(EditorPageContext);
+
+  const [editing, setEditing] = useState(false);
+
+  const [title, setTitle] = useState('');
+  useEffect(() => {
+    if (blockLoading === ActionStates.FINISHED) {
+      setTitle(blockValue ? blockValue.data.display_name : '');
```
connorhaugh, line 24: Ok so the default title will be '' if the xbock doesn't provide one. Is there a chance we could make this the type of the xblock? Does that make sense?

src/editors/EditorHeader.jsx
```diff
@@ -1,44 +1,112 @@
...
+  };
 
   const onCancelClicked = () => {
     if (unitUrlLoading === ActionStates.FINISHED) {
       const destination = `${studioEndpointUrl}/container/${unitUrl.data.ancestors[0].id}`;
       window.location.assign(destination);
     }
   };
+
+  let titleDisplay = 'Loading...';
```
connorhaugh, line 52: Why are we sometimes rendering a string, and sometimes rendering a react component here, depending on the loading state? I would caution that might not be great for the internationalization of the product. That can be easily fixed by replacing strings with formatted messages.

 Also, I'm not super comfortable with the  using a let here. It can get out of sync with the render in odd ways. I'll add a suggestion down below as to how we might do this behavior a little more cleanly.

src/editors/EditorHeader.test.jsx
```diff
@@ -48,10 +48,10 @@ test('Rendering And Click Button: Loaded Navigates Away', () => {
   };
   render(
     <EditorPageContext.Provider value={context}>
-      <EditorHeader title={title} />
+      <EditorHeader blockType={blockType} />
     </EditorPageContext.Provider>,
   );
-  expect(screen.getByText(title)).toBeTruthy();
+  // expect(screen.getByLabelText('Edit')).toBeTruthy();
```
connorhaugh, line 54: Why is this commented out? this seems like an important test, as we now need to know you can edit the title?

connorhaugh (changes requested):
Just a few questions, Otherwise looks really great!

src/editors/EditorHeader.jsx
connorhaugh: What behavior does onBlur accomplish, I'm not sure why Blur performs the tasks required when "I'm done editing"

rayzhou-bit: I intended for onChange and onKeyDown to update local state while onBlur updates the state in the provider. I think putting it all in onKeyDown makes sense. I'll update the code for this.

src/editors/EditorHeader.jsx
rayzhou-bit: Currently, when entering empty text into the text editor title, it will display "Editing: Component" and it will save as "html" (then when you reopen, it will display "Editing: html"). I think type of xblock would make sense here. Thoughts?

src/editors/EditorHeader.jsx
```diff
@@ -1,44 +1,112 @@
...
+    }
+  }
+
   return (
     <div className="editor-header">
       <ModalDialog.Header>
         <ActionRow>
           <ModalDialog.Title>
-            {mapBlockTypeToName(title)}
+            {titleDisplay}
```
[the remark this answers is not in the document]
rayzhou-bit: This sounds much neater than my current code. I will make this change.

src/editors/EditorHeader.test.jsx
rayzhou-bit: I commented this out because the edit button isn't there during the initial load (the 'Loading...' text is rendered without an edit button). In hindsight, I can make a test for when loading has completed.

src/editors/EditorHeader.jsx
```diff
@@ -1,44 +1,112 @@
...
-            aria-label="Cancel Changes and Return to Learning Context"
             src={Close}
             iconAs={Icon}
+            aria-label="Cancel Changes and Return to Learning Context"
             alt="Close"
             onClick={onCancelClicked}
-            variant="light"
             className="mr-2"
+            variant="light"
           />
```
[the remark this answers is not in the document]
rayzhou-bit: Alphabetical makes sense... I'll make changes so we follow this type of ordering.

src/editors/EditorHeader.jsx
rayzhou-bit: On re-examining this, onBlur covers the case of when the user clicks out of the edit.

connorhaugh: Ok, that makes sense I guess if it accomplishes the task.

src/editors/EditorHeader.jsx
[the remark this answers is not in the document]
connorhaugh: I will note that this is better, but you're still using a `let` variable, which can very easily get out of sync with the render. A const function as I suggested would ameliorate that issue. Lmk if that makes sense

src/editors/EditorHeader.test.jsx
connorhaugh: I would agree that we need tests for that kind of behavior. One thing that is helpful for me when deciding if I need a test is to think about writing down as many short sentences as you need to describe the behavior of the component. eg: `when loading, the component renders '...loading' and nothing else. When loaded, it renders the display, and the edit button. once you click the button, an input shows up.` Those sentences become your unit tests.

src/editors/EditorHeader.jsx
```diff
@@ -1,112 +1,124 @@
...
   } = useContext(EditorPageContext);
 
   const [editing, setEditing] = useState(false);
-
-  const [title, setTitle] = useState('');
+  const [title, setTitle] = useState(mapBlockTypeToName(blockType));
   useEffect(() => {
     if (blockLoading === ActionStates.FINISHED) {
-      setTitle(blockValue ? blockValue.data.display_name : '');
+      setTitle(blockValue ? blockValue.data.display_name : mapBlockTypeToName(blockType));
```
rayzhou-bit, line 25: Title now defaults to block type if it does not receive a value after loading.

src/editors/EditorHeader.jsx
```diff
@@ -1,112 +1,124 @@
...
   }, [blockLoading]);
 
-  const handleChange = (e) => {
-    setTitle(e.target.value);
+  const updateTitle = () => {
+    /* This is called when the input field loses focus.
+       It will update the title we are saving and set editing to false. */
+    setBlockTitle(title);
+    setEditing(false);
   };
```
rayzhou-bit, line 34: I messed around with this logic for a bit, but it makes the most sense to me to save and set editing to false when the input field loses focus.

src/editors/EditorHeader.jsx
```diff
@@ -1,44 +1,124 @@
...
+    );
+  };
+
   return (
     <div className="editor-header">
       <ModalDialog.Header>
         <ActionRow>
           <ModalDialog.Title>
-            {mapBlockTypeToName(title)}
+            {editableTitle(blockLoading !== ActionStates.FINISHED, editing)}
```
connorhaugh, line 102: Nice, this seems a lot cleaner, Thanks!

connorhaugh:
This looks good, thanks for taking my feedback into consideration. A couple of things before I thumb. I'm not sure why this is decreasing code coverage. Can you look into that? Also, you'll have to  resolve conficts with master branch before you merge.

rayzhou-bit:
Thanks for the feedback! I still do feel like this code can be improved so please feel free to suggest any more changes. I'll look into putting in some more testing. There's at least one more test case I can think of adding right now.

PR #18, opened 2022-02-22
chore: add tests for EditorHeader hooks

muselesscreator:
Separate hooks for the EditorHeader into reasonable testable chunks and test them individually :-).


connorhaugh (approved):
LGTM-- I implemented my hooks based on these and they seem to work!

PR #21, opened 2022-02-23
test: add TextEditor Tests

connorhaugh:
Imageuploadmodal will be added once imageupload is a real component.

### DIFF
--- a/src/editors/containers/TextEditor/TextEditor.test.jsx
+++ b/src/editors/containers/TextEditor/TextEditor.test.jsx
@@ -44,7 +44,7 @@ describe('TextEditor', () => {
   const props = {
     setEditorRef: jest.fn().mockName('args.setEditorRef'),
     // redux
-    blockValue: { data: 'HYleTsEditTeaxt' },
+    blockValue: { data: 'eDiTablE Text' },
     blockFailed: false,
     blockFinished: true,
     initializeEditor: jest.fn().mockName('args.intializeEditor'),
--- a/src/editors/containers/TextEditor/TextEditor.test.jsx
+++ b/src/editors/containers/TextEditor/TextEditor.test.jsx
@@ -3,7 +3,7 @@ import { shallow } from 'enzyme';
 import { TextEditor, mapStateToProps, mapDispatchToProps } from './TextEditor';
 import { actions, selectors } from '../../data/redux';
 import { RequestKeys } from '../../data/constants/requests';
-import {editorConfig, modalToggle} from './hooks';
+import { modalToggle } from './hooks';
 
 // Per https://github.com/tinymce/tinymce-react/issues/91 React unit testing in JSDOM is not supported by tinymce.
 // Consequently, mock the Editor out.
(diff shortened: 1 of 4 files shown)


src/editors/containers/TextEditor/hooks.test.jsx
```diff
@@ -0,0 +1,110 @@
...
+            addButton: mockAddbutton,
+          },
+        },
+      };
+      let output;
+      beforeEach(() => {
+        output = module.addImageUploadButton(mockOpenModal);
+      });
+      test('It calls addButton in the editor, but openModal is not called', () => {
+        output(editor);
```
rayzhou-bit, line 30: Should this be done with shallow()?

connorhaugh: Shallow is for testing rendering behavior-- we're not rendering anything here, as this is testing the behavior of a hook, not a component?

rayzhou-bit: I see. That makes sense.

src/editors/containers/TextEditor/TextEditor.test.jsx
```diff
@@ -1,51 +1,86 @@
...
 // Per https://github.com/tinymce/tinymce-react/issues/91 React unit testing in JSDOM is not supported by tinymce.
 // Consequently, mock the Editor out.
-const mockRole = 'Tiny-MCE-Mock';
 jest.mock('@tinymce/tinymce-react', () => {
   const originalModule = jest.requireActual('@tinymce/tinymce-react');
   return {
     __esModule: true,
     ...originalModule,
-    Editor: () => <div role={mockRole} />
+    Editor: () => <div />
```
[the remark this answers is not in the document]
connorhaugh: I changed it to be a string, but since this is a shallow render, it shouldn't show up anyway. But if you do see that string in the shallow render, something has messed up, so thats good.

src/editors/containers/TextEditor/TextEditor.test.jsx
```diff
@@ -1,51 +1,91 @@
...
+      isFinished: jest.fn((state, params) => ({ isFailed: { state, params } })),
+    },
+  },
+}));
+
+describe('TextEditor', () => {
+  const props = {
+    setEditorRef: jest.fn().mockName('args.setEditorRef'),
+    // redux
+    blockValue: { data: 'HYleTsEditTeaxt' },
```
muselesscreator, line 47: less brain-hurty plz? o.0

muselesscreator (commented):
My only request is a slightly less painful variable name ... 0.o

src/editors/containers/TextEditor/hooks.test.jsx
```diff
@@ -0,0 +1,110 @@
...
+      output.openModal();
+      expect(React.updateState).toHaveBeenCalledWith({ val: false, newVal: true });
+    });
+    test('returns closeModal field, which sets modal to true and calls updateState', () => {
+      output.closeModal();
+      expect(React.updateState).toHaveBeenCalledWith({ val: false, newVal: false });
+    });
+  });
+  describe('nullMethod hook', () => {
+    test('it outputs null', () => {
```
muselesscreator, line 106: I would omit this test.  that it doesn't return anything is not a good way to test that it doesn't do anything, and that is all we care about.

muselesscreator: (and there isn't a good way to test that it truly does nothing, just that it doesn't do other specific things.  But its being used as a callback here, and those pretty much always return undefined.

muselesscreator (approved):
;gtm

rayzhou-bit (approved):

PR #22, opened 2022-02-24
feat: image upload skeleton

muselesscreator:
Adds skeleton component with mock images and data flow for enabling work on both the ImageSelectionModal and the ImageSettingsModal.

Grabs the initial dimensions for the image because we can

https://user-images.githubusercontent.com/5533134/155615671-e1730f86-a374-4bf6-9c1e-53b4284f0d45.mov

.


connorhaugh:
After having thought about the setup over the weekend, I have an interesting question about this organization. One of the future tasks will be to make it so clicking on an image in the editor brings up the image settings page. I'm confused about how that input (html) will integrate into this flow? will the `selected` file be a javascript object with the html properties that we can feed into the form?

src/editors/containers/TextEditor/components/ImageSettingsModal/index.jsx
```diff
@@ -0,0 +1,147 @@
...
+  Button,
+  Form,
+  Image,
+} from '@edx/paragon';
+
+import BaseModal from '../BaseModal';
+import * as module from '.';
+
+export const hooks = {
+  dimensions: () => {
```
connorhaugh, line 13: We'll probaly want to have the initialization of these hooks take input, for the case in which the image settings are being edited.

src/editors/containers/TextEditor/components/ImageSettingsModal/index.jsx
```diff
@@ -0,0 +1,147 @@
...
+import {
+  Button,
+  Form,
+  Image,
+} from '@edx/paragon';
+
+import BaseModal from '../BaseModal';
+import * as module from '.';
+
+export const hooks = {
```
connorhaugh, line 12: I forget, are we separating these hooks out into there own file or have I converted you?

src/editors/data/redux/thunkActions/app.js
```diff
@@ -43,9 +44,15 @@ export const saveBlock = ({ content, returnToUnit }) => (dispatch) => {
   }));
 };
 
+export const fetchImages = ({ onSuccess }) => () => {
```
connorhaugh, line 47: We'll forgoe docstrings here for now, but when this hook gets plugged into the network we'll want to add them here-- just a note for me.

connorhaugh (approved):
I'm going to approve with the knowledge that we will iterate on this, and that it is in a good state for now.

src/editors/data/redux/thunkActions/app.js
muselesscreator: 100%

PR #24, opened 2022-02-25
Test  add requests redux tests

connorhaugh:
Add redux tests as well as get save functionality working.


src/editors/data/redux/requests/reducer.test.js
```diff
@@ -0,0 +1,32 @@
...
+      describe(action, () => {
+        it(`load ${target} from payload`, () => {
+          expect(reducer(testingState, actions[action](payload))).toEqual({
+            ...testingState,
+            [target]: testValue,
+          });
+        });
+      });
+    };
+    requestsList.flatMap(val => createTestParams(val)).map((value) => lifecycletest(...value));
```
muselesscreator, line 30: this is very concise, but a bit hard to read/parse...
Something like this may be easier to read, without un-necessarily returning an output array.
```
requestsList.forEach(requestKey => {
  createTestParams(requestKey).forEach((args) => lifecycleTest(...args));
});
```

src/editors/data/redux/requests/reducer.test.js
```diff
@@ -0,0 +1,32 @@
...
+  });
+  describe('handling actions', () => {
+    const requestsList = ['fetchUnit', 'fetchBlock', 'saveBlock'];
+    const createTestParams = (requestKey) => [
+      [requestKey, 'startRequest', requestKey, { status: RequestStates.pending }],
+      [requestKey, 'completeRequest', { requestKey }, { status: RequestStates.completed, response: undefined }],
+      [requestKey, 'failRequest', { requestKey }, { status: RequestStates.failed, error: undefined }],
+      [requestKey, 'clearRequest', { requestKey }, {}],
+    ];
+    const lifecycletest = (target, action, payload, testValue) => {
```
muselesscreator, line 20: I would give this a slightly clearer name (testLifecycleActions), and have the foreach list included HERE rather than in the createTestParams.

src/editors/data/redux/thunkActions/requests.test.js
```diff
@@ -1,15 +1,24 @@
...
-  lockSubmission: ({ submissionUUID }) => ({ lockSubmission: { submissionUUID } }),
-  unlockSubmission: ({ submissionUUID }) => ({ unlockSubmission: { submissionUUID } }),
-  updateGrade: (submissionUUID, gradeData) => ({ updateGrade: { submissionUUID, gradeData } }),
+const studioEndpointUrl = 'ASVouAGE.S$e';
+const blockId = 'CHalockIDmaiw@alcioUSness';
+const blockType = 'hTMl';
+const courseId = 'BEnX:INtrOToUNIttEsTing';
+const title = 'MYbLock';
+
+jest.spyOn(selectors.app, 'studioEndpointUrl').mockImplementation(() => studioEndpointUrl);
```
muselesscreator, line 12:
```suggestion
jest.mock('../app/selectors', () => ({
  studioEndpointUrl: (state) => ({ studioEndpointUrl: state }),
  blockId: (state) => ({ blockId: state }),
  blockType: (state) => ({ blockType: state }),
  courseId: (state) => ({ blockType: state }),
  blockTitle: (state) => ({ blockTitle: state }),
 }));
```

muselesscreator (changes requested):
some testing requests

src/editors/data/redux/requests/reducer.test.js
```diff
@@ -0,0 +1,48 @@
...
+  arbitraryField: 'arbitrary',
+};
+describe('requests reducer', () => {
+  test('intial state generated on create', () => {
+    expect(reducer(undefined, {})).toEqual(initialState);
+  });
+
+  describe('handling actions', () => {
+    const arbitraryKey = 'ArbItrAryKey';
+    const requestsList = [RequestKeys.fetchUnit, RequestKeys.fetchBlock, RequestKeys.saveBlock, arbitraryKey];
```
connorhaugh, line 15: @muselesscreator here's the arbitrary key!

src/editors/data/redux/thunkActions/requests.test.js
```diff
@@ -1,15 +1,26 @@
...
-  unlockSubmission: ({ submissionUUID }) => ({ unlockSubmission: { submissionUUID } }),
-  updateGrade: (submissionUUID, gradeData) => ({ updateGrade: { submissionUUID, gradeData } }),
+const studioEndpointUrl = 'ASVouAGE.S$e';
+const blockId = 'CHalockIDmaiw@alcioUSness';
+const blockType = 'hTMl';
+const courseId = 'BEnX:INtrOToUNIttEsTing';
+const title = 'MYbLock';
+
+jest.mock('../app/selectors', () => ({
+  studioEndpointUrl: () => (studioEndpointUrl),
```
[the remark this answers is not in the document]
muselesscreator:
```
myTestSelector: (state) => ({ myTestSelector: state }),
myComplicatedSelector: (state, props) => ({ myComplicatedSelector: { state, props } }),
```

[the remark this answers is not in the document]
connorhaugh: I don't get this, because the selectors we are mocking don't output the whole state, and why would we mock something with different than desired behavior?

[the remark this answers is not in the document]
connorhaugh: Do you mean something like? `myTestSelector: (state) => ({ state[myTestSelector]})` because that would be closer to what a selector is, but I'm not sure I like that either?

src/editors/data/redux/thunkActions/requests.test.js
```diff
@@ -101,77 +113,64 @@ describe('requests thunkActions module', () => {
       });
     });
   };
-
   describe('network request actions', () => {
-    const submissionUUID = 'test-submission-id';
-    const locationId = 'test-location-id';
+    const sometestStateGetter = () => ({
```
muselesscreator, line 117: this should really just be:
```
const getState = () => testState;
```
then, each of your tests against selectors just check against the object, not by calling the getter again.

src/editors/data/redux/thunkActions/requests.test.js
```diff
@@ -78,12 +89,13 @@ describe('requests thunkActions module', () => {
...
     action,
     args,
+    testState,
     expectedData,
     expectedString,
   }) => {
     let dispatchedAction;
     beforeEach(() => {
-      action({ ...args, onSuccess, onFailure })(dispatch);
+      action({ ...args, onSuccess, onFailure })(dispatch, testState);
```
muselesscreator, line 98:
```suggestion
      action({ ...args, onSuccess, onFailure })(dispatch, () => testState);
```
and should use the top-level `testState` object, not one passed in here

src/editors/data/redux/thunkActions/requests.test.js
```diff
@@ -101,77 +113,64 @@ describe('requests thunkActions module', () => {
...
+        expectedString: 'with fetchBlock promise',
         expectedData: {
-          requestKey: RequestKeys.setLock,
-          promise: api.lockSubmission(submissionUUID),
+          ...sometestStateGetter(),
+          requestKey: RequestKeys.fetchBlock,
+          promise: api.fetchBlockById({
+            studioEndpointUrl: selectors.app.studioEndpointUrl(sometestStateGetter()),
+            blockId: selectors.app.blockId(sometestStateGetter()),
+          }),
```
muselesscreator, line 136: The args need to be different, as they are specifically fetch parameters.
With the above changes, this becomes:
```
      const fetchParams = { fetchParam1: 'param1', fetchParam2: 'param2' };
      testNetworkRequestAction({
        action: requests.fetchBlock,
        args: fetchParams,
        expectedString: 'with fetchBlock promise',
        expectedData: {
          ...fetchParams,
          requestKey: RequestKeys.fetchBlock,
          promise: api.fetchBlockById({
            studioEndpointUrl: selectors.app.studioEndpointUrl(testState),
            blockId: selectors.app.blockId(testState),
          }),
        },
      });
```

connorhaugh:
Ready for another round of review at your earliest convenience @muselesscreator

src/editors/data/redux/requests/reducer.test.js
```diff
@@ -0,0 +1,48 @@
...
+        });
+        test('failRequest sets failed state and loads error', () => {
+          testAction(
+            'failRequest',
+            { requestKey },
+            { status: RequestStates.failed },
+          );
+        });
+        test('clearRequest clears request state', () => {
+          testAction('clearRequest', { requestKey }, {});
```
muselesscreator, line 43: For this test, you should probably specifically give an input test state that has a non-empty state for the given request.  If you want, you could achieve this in your testAction method by calling the reducer with an initial state of ({ ...testingState, [requestKey]: { arbitrary: 'state' } });

Reasoning here is that we are importing the initial state, and this is a reasonable initial state to have expected.

src/editors/data/redux/thunkActions/requests.test.js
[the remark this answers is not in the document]
muselesscreator: We discussed this offline, but in case anyone comes here looking for the answer later:
We are mocking the selectors here to give a return value that includes an identification of which selector was called, and how it was called, in order to avoid having to test those pieces independently, and rather than simply checking that we get the output of a selector when called with *any* arguments.

It is much more important in **unit tests** to verify what the code is specifically trying to do, rather than to perfectly model the use case it is intended to serve.

connorhaugh:
@muselesscreator should be all set to go here.

muselesscreator (approved):
lgtm

PR #27, opened 2022-03-02
Gallery app

muselesscreator:
Adds an offline/Gallery view for the editors.
Accessible by navigating to `www/` and running `npm install` followed by `npm start`.


rayzhou-bit (approved):
this looks good to me

connorhaugh:
Did this not run CI checks for a reason? Or did this change mess things up?

www/src/Gallery.jsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+
+import { Form } from '@edx/paragon';
+
+import { EditorPage } from '@edx/frontend-lib-content-components';
+
+export const EditorGallery = () => {
+  const [blockType, setBlockType] = React.useState('html');
+  const blockIds = {
```
connorhaugh, line 9: This 1. should be imported from the file where we specify the block types. 2 should be call `blockTypes` not `blockIds`

connorhaugh, line 9: this should be an dict with keys and values generated by the source of truth of block types available, which right now is src/frontend-lib-content-components/src/editors/data/constants/app.js. That way we don't have to update this view every time someone adds an editor

www/src/i18n/index.js
```diff
@@ -0,0 +1,18 @@
+import arMessages from './messages/ar.json';
+// no need to import en messages-- they are in the defaultMessage field
+import es419Messages from './messages/es_419.json';
+import frMessages from './messages/fr.json';
+import kokrMessages from './messages/ko_kr.json';
+import ptbrMessages from './messages/pt_br.json';
+import zhcnMessages from './messages/zh_CN.json';
+
+const messages = {
```
connorhaugh, line 9: Will this override languages when we release the npm package to the piblic?

muselesscreator: This is just the gallery.

muselesscreator: well... specifically, this is generating the intl top-level provider that enables the subcomponents to also have intl capability (which they require)

www/module.config.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
```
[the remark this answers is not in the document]
muselesscreator: honestly, because webpack is weird, and even though this seems like it should do exactly the same thing as the webpack config, it fails without this file...

www/src/Gallery.jsx
muselesscreator: Fair.  was trying to avoid tying them too closely together, but that is probably a bit silly in this case.

connorhaugh (approved):
given that this now runs tests, I'm more comfortable thumbing it, thanks for doing those little nits, it will make this easier for making this an openedx ready product. I'm now realizing we had a github actions blackout sometime last week maybe which might have caused the tests to not run?

PR #30, opened 2022-03-15
feat: Image settings feature and design completeness

muselesscreator:
image

The current dimension lock behavior should be stable, but may not be optimal.  A better user experience would need a solid design though.

### DIFF
--- a/src/editors/containers/TextEditor/components/ImageSettingsModal/AltTextControls.jsx
+++ b/src/editors/containers/TextEditor/components/ImageSettingsModal/AltTextControls.jsx
@@ -1,0 +1,48 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Form } from '@edx/paragon';
+
+import * as hooks from './hooks';
+
+/**
+ * Wrapper for alt-text input and isDecorative checkbox control
+ * @param {bool} isDecorative - is the image decorative?
+ * @param {func} setIsDecorative - handle isDecorative change event
+ * @param {func} setValue - update alt-text value
+ * @param {string} value - current alt-text value
+ */
+export const AltTextControls = ({
+  isDecorative,
+  setIsDecorative,
+  setValue,
+  value,
+}) => (
+  <Form.Group className="mt-4.5">
+    <Form.Label as="h4">Accessibility</Form.Label>
+    <Form.Control
+      className="mt-4.5"
+      type="input"
+      value={value}
+      disabled={isDecorative}
+      onChange={hooks.onInputChange(setValue)}
+      floatingLabel="Alt Text"
+    />
+    <Form.Checkbox
+      className="mt-4.5 decorative-control-label"
+      checked={isDecorative}
+      onChange={hooks.onCheckboxChange(setIsDecorative)}
+    >
+      <Form.Label>
+        This image is decorative (no alt text required).
+      </Form.Label>
+    </Form.Checkbox>
+  </Form.Group>
+);
+AltTextControls.propTypes = {
+  isDecorative: PropTypes.bool.isRequired,
+  value: PropTypes.string.isRequired,
+  setValue: PropTypes.func.isRequired,
+  setIsDecorative: PropTypes.func.isRequired,
+};
+
+export default AltTextControls;
--- a/src/editors/containers/TextEditor/components/ImageSettingsModal/hooks.js
+++ b/src/editors/containers/TextEditor/components/ImageSettingsModal/hooks.js
@@ -1,0 +1,251 @@
+import React from 'react';
+
+import { StrictDict } from '../../../../utils';
+import * as module from './hooks';
+
+// Simple wrappers for useState to allow easy mocking for tests.
+export const state = {
+  dimensions: (val) => React.useState(val),
+  locked: (val) => React.useState(val),
+  local: (val) => React.useState(val),
+  lockInitialized: (val) => React.useState(val),
+  altText: (val) => React.useState(val),
+  isDecorative: (val) => React.useState(val),
+};
+
+export const dimKeys = StrictDict({
+  height: 'height',
+  width: 'width',
+});
+
+/**
+ * findGcd(numerator, denominator)
+ * Find the greatest common denominator of a ratio or fraction.
+ * @param {number} numerator - ratio numerator
+ * @param {number} denominator - ratio denominator
+ * @return {number} - ratio greatest common denominator
+ */
+export const findGcd = (a, b) => (b ? findGcd(b, a % b) : a);
+const checkEqual = (d1, d2) => (d1.height === d2.height && d1.width === d2.width);
+
+/**
+ * getValidDimensions({ dimensions, local, locked })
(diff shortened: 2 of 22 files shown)


src/editors/containers/TextEditor/components/ImageSettingsModal/hooks.js
```diff
@@ -0,0 +1,251 @@
...
+ *   {func} lock - lock current dimensions
+ *   {func} unlock - unlock dimensions
+ *   {object} value - current dimension values
+ *   {func} setHeight - set height
+ *     @param {string} - new height string
+ *   {func} setWidth - set width
+ *     @param {string} - new width string
+ *   {func} updateDimensions - set dimensions based on state
+ */
+export const dimensionHooks = () => {
```
connorhaugh, line 146: This will have to initialize with either null, or some value (from image selected from editor, as we need to respect the html). Shouldn't be too much of a change, but wanted to point out this  (+ same for alt text) is where our two PRs collide.

src/editors/containers/TextEditor/components/ImageSettingsModal/hooks.js
```diff
@@ -0,0 +1,251 @@
...
+    setLocal(value);
+  };
+  const {
+    initializeLock,
+    lock,
+    locked,
+    unlock,
+  } = module.dimensionLockHooks({ dimensions });
+  return {
+    onImgLoad: (selection) => ({ target: img }) => {
```
connorhaugh, line 160: or it seems we are doing that init value here only on image Load?

src/editors/containers/TextEditor/components/ImageSettingsModal/AltTextControls.jsx
```diff
@@ -0,0 +1,48 @@
...
+  <Form.Group className="mt-4.5">
+    <Form.Label as="h4">Accessibility</Form.Label>
+    <Form.Control
+      className="mt-4.5"
+      type="input"
+      value={value}
+      disabled={isDecorative}
+      onChange={hooks.onInputChange(setValue)}
+      floatingLabel="Alt Text"
+    />
```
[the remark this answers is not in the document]
connorhaugh: So my question here, better phrased is this:

Is disabling the save button a sufficient amount of feedback to give to the user that they need to either set the image to be decorative or add alt text? To me, It doesn't, it just looks like something is brokem. I think it would be prudent to add some feedback to the user that they need to fill in either of these. The `Form.Control.Feedback` component in paragon might be helpful to do that. If this is future work, I'm more than happy to write a ticket, but this feels important enough to include in this.

src/editors/containers/TextEditor/components/ImageSettingsModal/hooks.js
```diff
@@ -0,0 +1,251 @@
...
+ *   {obj} locked - current locked state
+ *   {func} lock - lock the current dimensions
+ *   {func} unlock - unlock the dimensions
+ */
+export const dimensionLockHooks = ({ dimensions }) => {
+  const [locked, setLocked] = module.state.locked(null);
+  const [lockInitialized, setLockInitialized] = module.state.lockInitialized(null);
+  const lock = () => module.lockDimensions({ lockInitialized, dimensions, setLocked });
+
+  React.useEffect(lock, [lockInitialized]);
```
connorhaugh, line 119: Just to clarify-- here, whenever `lockInitialized` changes, the lock function is called to set the dimensions to whatever they were the first time the lock was initialized, or is dimensionLockHooks also called every render, so that this value would lock to the current dimensions. I know it is most likely the former, but want to be clear on our approach.

src/editors/containers/TextEditor/components/ImageSettingsModal/hooks.test.js
```diff
@@ -0,0 +1,314 @@
...
+        expect(setState.lockInitialized).toHaveBeenCalledWith(true);
+      });
+      test('lock is successful if lockInitialized is true', () => {
+        mockStateVal(stateKeys.lockInitialized, true, setState.lockInitialized);
+        hook = hooks.dimensionLockHooks({ dimensions: simpleDims });
+        expect(hook.locked).toEqual(null);
+        hook.lock();
+        expect(setState.locked).toHaveBeenCalledWith({ ...simpleDims });
+      });
+      test('unlock sets locked to null', () => {
```
connorhaugh, line 184: Any reason we're doing locked: true unlocked: null instead of locked: true unlocked: false. Seems we're not bound to that convention, and it seems a little unintuitive?

src/editors/containers/TextEditor/components/ImageSettingsModal/hooks.test.js
```diff
@@ -0,0 +1,314 @@
...
+));
+
+let oldState;
+const setState = {};
+const mockState = () => {
+  oldState = hooks.state;
+  const keys = Object.keys(stateKeys);
+  keys.forEach(key => {
+    setState[key] = jest.fn();
+    hooks.state[key] = jest.fn(val => [val, setState[key]]);
```
connorhaugh, line 34: This step is critical, in that we are monkeypatching in our own value for hooks.state for each key, and because we are doing it multiple times, we don't mock at the module level? Just trying to clarify my understanding, so that I can identify places to use this pattern because it seems neat.

connorhaugh (commented):
So I think this looks great, and other than my question about form feedback, which either is a quick add or a follow up ticket, I just left some clarification questions around the hooks testing so I could use some of those patterns in future.

src/editors/containers/TextEditor/components/ImageSettingsModal/AltTextControls.jsx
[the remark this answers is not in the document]
muselesscreator: For right now, I would call that a new ticket, as the behavior is not spelled out in the design, and where/how we would want that feedback to be indicated is a design decision :-)

src/editors/containers/TextEditor/components/ImageSettingsModal/hooks.js
muselesscreator: whenever lockInitialized changes, it locks to the "current" dimensions, iff (if-and-only-if) the lock is initialized.
So, in practice, it gets run once, when the lock initializes the first time (and the component renders) and loads the initial dimension values.

src/editors/containers/TextEditor/components/ImageSettingsModal/hooks.test.js
muselesscreator: because locked is an object, and I want to test against truthiness not actual "===false".

connorhaugh: That makes sense.

src/editors/containers/TextEditor/components/ImageSettingsModal/hooks.js
connorhaugh: Ah ok, that makes sense. Thanks for all the additional clarification.

PR #36, opened 2022-03-21
Chore: Test coverage hunt

muselesscreator:
updating and improving test coverage across the app.

Added a simple testGetter method to the MockUseState util to make it a bit more helpful.


connorhaugh:
looks good so far 👍, going to wait till #34 is merged, and this passes lint/tests before giving it another look.

.gitignore
```diff
@@ -108,3 +108,5 @@ dist
 *.swo
 *.swp
 
+### local overrides ###
+module.config.js
```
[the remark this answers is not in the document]
muselesscreator: brand.

src/editors/containers/TextEditor/components/ImageSettingsModal/DimensionControls.jsx
```diff
@@ -23,7 +23,7 @@ import hooks from './hooks';
  * @param {obj} value - local dimension values { height, width }
  */
 export const DimensionControls = ({
-  locked,
+  isLocked,
```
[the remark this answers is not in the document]
muselesscreator: yup

src/editors/containers/TextEditor/components/ImageUploadModal.jsx
```diff
@@ -1,44 +1,67 @@
...
+import tinyMCEKeys from '../../../data/constants/tinyMCE';
 import ImageSettingsModal from './ImageSettingsModal';
 import SelectImageModal from './SelectImageModal';
 import * as module from './ImageUploadModal';
 
+export const propsString = (props) => Object.keys(props)
+  .map(key => `${key}="${props[key]}"`)
+  .join(' ');
+
+export const imgProps = ({ settings, selection }) => ({
```
connorhaugh, line 13: nit of the most nitpicky (of work that I think is partially mine lol): Is it potentially misleading to call something which is not React  a `prop` instead of an `attribute`

package.json
```diff
@@ -40,11 +40,13 @@
     "@testing-library/dom": "^8.11.1",
     "@testing-library/react": "12.1.1",
     "@testing-library/user-event": "^13.5.0",
+    "axios": "^0.26.1",
```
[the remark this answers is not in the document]
muselesscreator: I honestly have no idea... might be because we weren't explicitly importing it and an `npm install` mucked with the registry.

src/editors/containers/TextEditor/components/ImageUploadModal.jsx
muselesscreator: eh, I wouldn't say so.  Is a react app and we are building UI html references.  I would almost rather we be using JSX with a react method to use JSX here rather than a flat string.

```
const htmlString = ReactDOMServer.renderToStaticMarkup(
      <div ...
);
```

connorhaugh: Not saying we have to do it but just want to follow your thinking:  Would the advantage of using `renderToStaticMarkup` that it would be more readable because jsx gives us nice non-string format and strings are harder to read than code? Or is there a performance// maintainability thing here that i'm missing?

www/module.config.js
```diff
@@ -7,6 +7,7 @@ module.exports = {
     dist: The sub-directory of the source code where it puts its build artifact.  Often "dist".
```
connorhaugh, line 7: do we need to gitignore this one too? Seems like it won't work unless you are edx

connorhaugh (approved):
Looks good, just one file in www to gitignore and follow up question to a nit we don't need to act on. Test coverage looks great.

src/editors/containers/TextEditor/components/ImageUploadModal.jsx
muselesscreator: this is less verifyable.  Is more "read this string... does it look right?" whereas jsx has built-in validation that we are making a valid html element

www/module.config.js
muselesscreator: uuuuuugh. so, we need the part that links to the platform repo.  I think this is a place where we should use this to alias to another locally-checked-out repo in the www package.json.  I will poke at.

PR #41, opened 2022-03-25
Select img api

muselesscreator:
Select Image api updates from Connor's PR, adding tests.

A couple more I would like to add in the api for the new image ingestion transform, as well as in utils for the sort methods, but this gets us almost all the way there.


src/editors/containers/TextEditor/components/SelectImageModal/hooks.js
```diff
@@ -31,15 +34,15 @@ export const displayList = ({ sortBy, searchString, images }) => module.filtered
...
-  fetchImages,
   setSelection,
   searchSortProps,
 }) => {
+  const dispatch = useDispatch();
   const [images, setImages] = module.state.images({});
   const [highlighted, setHighlighted] = module.state.highlighted(null);
 
   React.useEffect(() => {
-    fetchImages({ onSuccess: setImages });
```
rayzhou-bit, line 42: We're getting error "checkPropTypes.jss?a15c:20 Warning: Failed prop type: The prop `highlighted` is marked as required in `Gallery`, but its value is `null`."

Should we change the PropTypes check or init with ""?

connorhaugh (approved):
At this point, I am comfortable with thumbing this, and doing the fast-follow work to make image upload calls work correctly, improve test coverage, and set the overflow scroll. I have certainly exercised every other part of the workflow this touches.

PR #42, opened 2022-03-25
fix: Allow latest frontend-platform versions

Jawayria:
Updated version constraint on frontend-platform to allow latest versions of the package to be installed.


connorhaugh:
closed, as this is now out of date.

PR #43, opened 2022-03-30
Feat: Add Request Alerts and upload file

connorhaugh:
Fixes to Upload data type, as well as adding in of two error alerts for upload and fetch.


kenclary (approved):
looks alright to me. and we'll have opportunity to iterate later, if needed.

rayzhou-bit:
Should FetchErrorAlert and UploadErrorAlert be the same file? Everything else looks good to me.

connorhaugh:
> Should FetchErrorAlert and UploadErrorAlert be the same file? 
Great question. Off the top of my head They honestly could be the same component, which takes in a prop "which request key do I look at and what message do I take." If you think that refactor is worthwhile, lmk (it probably is tbh)

rayzhou-bit:
> > Should FetchErrorAlert and UploadErrorAlert be the same file?
> 
> Great question. Off the top of my head They honestly could be the same component, which takes in a prop "which request key do I look at and what message do I take." If you think that refactor is worthwhile, lmk (it probably is tbh)

This would be worth it for video editor work and for if we expand errors in the future. If time if an issue, could we submit a ticket to change this later?

connorhaugh:
> This would be worth it for video editor work and for if we expand errors in the future. If time if an issue, could we submit a ticket to change this later?

So In thinking about this more, I was wondering if the present pattern is the way to go, because:
1. error sources can come from more than requests, so having this built out as a pattern for multiple use cases makes sense, so for instance, if someone doesn't use redux to get error state, they know the base pattern is to use the error alert component.
2. This is quite simple and easy to read.

rayzhou-bit:
> So In thinking about this more, I was wondering if the present pattern is the way to go, because:
> 
> 1. error sources can come from more than requests, so having this built out as a pattern for multiple use cases makes sense, so for instance, if someone doesn't use redux to get error state, they know the base pattern is to use the error alert component.
> 2. This is quite simple and easy to read.

The readability of this is definitely very useful. I looked through the code more closely and I can see why you did it this way now.

rayzhou-bit (approved):

PR #44, opened 2022-03-31
Add events sent to Segment for the save and cancel clicks for content editors.

doctoryes:
https://openedx.atlassian.net/browse/TNL-9726

For the new React-based content editors, we'd like to know how many content creators are clicking the Cancel and Add To Course (Save) buttons in each editor. This PR adds an event for each click.

### DIFF
--- a/src/editors/analytics.js
+++ b/src/editors/analytics.js
@@ -1,0 +1,6 @@
+export const analytics = {
+  editorSaveClick: 'edx.ui.authoring.editor.save',
+  editorCancelClick: 'edx.ui.authoring.editor.cancel',
+};
+
+export default analytics;
--- a/src/editors/analyticsEvt.js
+++ b/src/editors/analyticsEvt.js
@@ -1,8 +1,6 @@
 export const analyticEvt = {
-  footer: {
-    editorSaveClick: 'edx.ui.authoring.editor.save',
-    editorCancelClick: 'edx.ui.authoring.editor.cancel',
-  },
+  editorSaveClick: 'edx.ui.authoring.editor.save',
+  editorCancelClick: 'edx.ui.authoring.editor.cancel',
 };
 
 export default analyticEvt;
--- a/src/editors/analyticsEvt.js
+++ b/src/editors/analyticsEvt.js
@@ -1,6 +1,6 @@
-export const analytics = {
+export const footerAnalytics = {
   editorSaveClick: 'edx.ui.authoring.editor.save',
   editorCancelClick: 'edx.ui.authoring.editor.cancel',
 };
 
-export default analytics;
+export default footerAnalytics;
--- a/src/editors/analyticsEvt.js
+++ b/src/editors/analyticsEvt.js
@@ -1,6 +1,8 @@
-export const footerAnalytics = {
-  editorSaveClick: 'edx.ui.authoring.editor.save',
-  editorCancelClick: 'edx.ui.authoring.editor.cancel',
+export const analyticEvt = {
+  footer: {
+    editorSaveClick: 'edx.ui.authoring.editor.save',
+    editorCancelClick: 'edx.ui.authoring.editor.cancel',
+  },
 };
 
-export default footerAnalytics;
+export default analyticEvt;
--- a/src/editors/containers/EditorContainer/components/EditorFooter/index.jsx
+++ b/src/editors/containers/EditorContainer/components/EditorFooter/index.jsx
@@ -28,7 +28,7 @@ export const EditorFooter = ({
       </Toast>
     )}
 
-    <ModalDialog.Footer>
+    <ModalDialog.Footer className="border-top-0">
       <ActionRow>
         <ActionRow.Spacer />
         <Button
--- a/src/editors/data/redux/app/selectors.js
+++ b/src/editors/data/redux/app/selectors.js
@@ -61,7 +61,7 @@ export const analytics = createSelector(
     module.simpleSelectors.courseId,
   ],
   (blockId, blockType, courseId) => (
-    {blockId, blockType, courseId}
+    { blockId, blockType, courseId }
   ),
 );
 
--- a/src/editors/data/redux/app/selectors.js
+++ b/src/editors/data/redux/app/selectors.js
@@ -54,14 +54,15 @@ export const displayTitle = createSelector(
       : blockType[0].toUpperCase() + blockType.substring(1);
   },
 );
-export const returnAnalytics = createSelector(
+export const analytics = createSelector(
   [
     module.simpleSelectors.blockId,
     module.simpleSelectors.blockType,
     module.simpleSelectors.courseId,
   ],
-  // (blockId, blockType, courseId) => {blockId, blockType, courseId},
-  (courseId) => (courseId),
+  (blockId, blockType, courseId) => (
+    {blockId, blockType, courseId}
+  ),
 );
 
 export default {
@@ -69,5 +70,5 @@ export default {
   isInitialized,
   returnUrl,
   displayTitle,
-  returnAnalytics,
+  analytics,
 };
--- a/src/editors/data/redux/app/selectors.js
+++ b/src/editors/data/redux/app/selectors.js
@@ -54,10 +54,20 @@ export const displayTitle = createSelector(
       : blockType[0].toUpperCase() + blockType.substring(1);
   },
 );
+export const returnAnalytics = createSelector(
+  [
+    module.simpleSelectors.blockId,
+    module.simpleSelectors.blockType,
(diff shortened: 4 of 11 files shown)


src/editors/components/EditorFooter/index.jsx
```diff
@@ -18,8 +19,15 @@ import { selectors, thunkActions } from '../../data/redux';
 import messages from './messages';
 import * as module from '.';
 
-export const handleSaveClicked = (props) => () => saveBlock(props);
-export const handleCancelClicked = ({ returnUrl }) => navigateCallback(returnUrl);
+export const handleSaveClicked = (props) => () => {
+  sendTrackEvent('edx.ui.authoring.editor.save', {});
```
doctoryes, line 23: Data bits that would be nice to send along with this event (and the one below):
- Block ID
- Block type
- Course run ID (might be unavailable)

Reviewers, how would I access this info to send it to the handlers for sending to Segment?

connorhaugh: You could have that information passed in from the redux state using mapStateToProps.

src/editors/components/EditorFooter/index.jsx
```diff
@@ -18,8 +19,15 @@ import { selectors, thunkActions } from '../../data/redux';
...
+  saveBlock(props);
+};
+
+export const handleCancelClicked = ({ returnUrl }) => {
+  sendTrackEvent('edx.ui.authoring.editor.cancel', {});
```
connorhaugh, line 28: So one thing we have been doing to avoid magic strings like this is having a javascript object stored in utils which takes as values the various ids of tracking events. Where this breaks down is that we want to have tracking events in the MFE and in this package, and passing them down into the editor as props seems unnecessary becuase we don't do the same for formatted message ids. I am wonder what you thoughts might be here @muselesscreator

src/editors/components/EditorFooter/index.jsx
connorhaugh: Does this make sense?

rayzhou-bit: I believe we should create some sort of returnAnalytics selector in /data/redux/app/selectors.js to get these values.

The sendTrackEvent call should be placed in hooks.js since this is where we've been writing our logics. We should try to keep the EditorFooter component as 'dumb' as possible.

src/editors/components/EditorFooter/index.jsx
doctoryes: Value exists in simply having the event name listed at the exact place it's emitted. Also, this pattern seems to be a standard used elsewhere: https://github.com/openedx/frontend-app-learning/search?q=sendTrackEvent

src/editors/components/EditorFooter/index.jsx
doctoryes: @connorhaugh @rayzhou-bit I've attempted to incorporate your feedback - but I've been unable to make a commit that passes tests, mostly because of my lack of Redux knowledge. This task will be completed much more quickly if one of you takes over this PR and makes the needed changes - so I recommend you talk amongst yourselves and decide who will complete it. Please let me know your decision - thanks.

src/editors/analytics.js
```diff
@@ -0,0 +1,6 @@
+export const analytics = {
```
muselesscreator, line 1: this should probably be in editors/data/constants

src/editors/components/EditorFooter/index.jsx
```diff
@@ -9,24 +9,38 @@ import {
...
-export const handleCancelClicked = ({ returnUrl }) => navigateCallback(returnUrl);
+export const handleSaveClicked = ({ saveProps, returnAnalytics }) => () => {
+  console.log('test save', returnAnalytics);
+  console.log('test save', sendTrackEvent)
+  sendTrackEvent(analytics.editorSaveClick, returnAnalytics);
+  saveBlock(saveProps);
+};
+
+export const handleCancelClicked = ({ returnUrl, returnAnalytics }) => {
+  console.log('test cancel', returnAnalytics);
```
muselesscreator, line 31: should remove

src/editors/components/EditorFooter/index.jsx
```diff
@@ -9,24 +9,38 @@ import {
...
 
 import messages from './messages';
+import analytics from '../../analytics';
 import * as module from '.';
 
-export const handleSaveClicked = (props) => () => saveBlock(props);
-export const handleCancelClicked = ({ returnUrl }) => navigateCallback(returnUrl);
+export const handleSaveClicked = ({ saveProps, returnAnalytics }) => () => {
+  console.log('test save', returnAnalytics);
+  console.log('test save', sendTrackEvent)
```
muselesscreator, line 25: should remove

muselesscreator: Also, these should be added into saveBlock and and navigateCallback in the editor hooks.js file

src/editors/components/EditorFooter/index.test.jsx
```diff
@@ -68,14 +68,14 @@ describe('EditorFooter', () => {
       createdCallback();
       expect(saveBlock).toHaveBeenCalled();
       expect(sendTrackEvent).toHaveBeenCalledTimes(1);
-      expect(sendTrackEvent).toHaveBeenCalledWith(analytics.editorSaveClick, props.analytics);
+      expect(sendTrackEvent).toHaveBeenCalledWith(analyticsEvt.footer.editorSaveClick, props.analytics);
     });
     test('handleCancelClicked calls navigateCallback', () => {
       sendTrackEvent.mockClear();
       realmodule.handleCancelClicked({ returnUrl: props.returnUrl });
       expect(navigateCallback).toHaveBeenCalledWith(props.returnUrl);
```
rayzhou-bit, line 76: The current code does not pass this test as this is a callback within a callback. This will be adjusted when adjusting to Ben's changes for the video editor.

rayzhou-bit:
I will code this into Ben's change for the Video Editor once that has merged.

src/editors/containers/EditorContainer/components/EditorFooter/index.jsx
```diff
@@ -28,7 +28,7 @@ export const EditorFooter = ({
       </Toast>
     )}
 
-    <ModalDialog.Footer>
+    <ModalDialog.Footer className="border-top-0">
```
rayzhou-bit, line 31: Visual fix for this line above cancel and save footer buttons.

src/editors/hooks.js
```diff
@@ -13,16 +15,31 @@ export const navigateTo = (destination) => {
...
 
-export const navigateCallback = (destination) => () => {
+export const navigateCallback = ({
+  destination,
+  analyticsEvent,
+  analytics,
+}) => () => {
+  if (process.env.NODE_ENV !== 'development' && analyticsEvent && analytics) {
+    sendTrackEvent(analyticsEvent, analytics);
+  }
```
rayzhou-bit, line 25: This is where the analytics is being sent. Effectively, it will happen when we navigate away from _any_ editor modal if analyticsEvent and analytics are passed to the call.

src/editors/data/redux/app/selectors.js
```diff
@@ -61,7 +61,7 @@ export const analytics = createSelector(
     module.simpleSelectors.courseId,
   ],
   (blockId, blockType, courseId) => (
-    {blockId, blockType, courseId}
+    { blockId, blockType, courseId }
```
rayzhou-bit, line 64: The analytics we want to send is set up here. This can be changed in the future if we want more than just blockId, blockType and courseId.

src/editors/analyticsEvt.js
```diff
@@ -0,0 +1,6 @@
+export const analyticEvt = {
```
connorhaugh, line 1: Might be worth putting this under data/constants/, in order to have it near all our other constants, but I'm open to whatever.

src/editors/containers/EditorContainer/components/EditorFooter/index.jsx
connorhaugh: Is the border line not supposed to be there in the hi-fi mocks? Just curious

src/editors/data/redux/app/selectors.js
connorhaugh: Maybe rename this to analyticsData or something that better describes that this is a payload to be used, not the actual analytics system itself? does that make sense?

src/editors/analyticsEvt.js
rayzhou-bit: I was modeling it after messages.jsx, but I like this more! Will move.

connorhaugh (approved):

PR #46, opened 2022-04-02
Feat: empty gallery

rayzhou-bit:
Added 2 labels for when gallery is empty:

1. When no images are returned from backend.
2. When search returns 0 results.

This does not match the mockups in figma and is more of a quick temporary solution.

### DIFF
--- a/src/editors/containers/TextEditor/components/SelectImageModal/Gallery.jsx
+++ b/src/editors/containers/TextEditor/components/SelectImageModal/Gallery.jsx
@@ -15,7 +15,8 @@ import messages from './messages';
 import GalleryCard from './GalleryCard';
 
 export const Gallery = ({
-  isEmpty,
+  galleryIsEmpty,
+  searchIsEmpty,
   displayList,
   highlighted,
   onHighlightChange,
@@ -33,10 +34,17 @@ export const Gallery = ({
       />
     );
   }
-  if (isEmpty) {
+  if (galleryIsEmpty) {
     return (
       <div className="gallery p-4 bg-gray-100" style={{ height: '375px' }}>
         <FormattedMessage {...messages.emptyGalleryLabel} />
+      </div>
+    );
+  }
+  if (searchIsEmpty) {
+    return (
+      <div className="gallery p-4 bg-gray-100" style={{ height: '375px' }}>
+        <FormattedMessage {...messages.emptySearchLabel} />
       </div>
     );
   }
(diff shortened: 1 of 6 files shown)


src/editors/containers/TextEditor/components/SelectImageModal/hooks.js
```diff
@@ -53,9 +54,11 @@ export const imgListHooks = ({
       onClick: () => setSelection(images[highlighted]),
     },
     galleryProps: {
+      galleryIsEmpty: Object.keys(images).length === 0,
```
connorhaugh, line 57: This is run once-- should it live inside a useffect, so it occurs on render?

src/editors/containers/TextEditor/components/SelectImageModal/Gallery.jsx
```diff
@@ -32,6 +34,20 @@ export const Gallery = ({
       />
     );
   }
+  if (galleryIsEmpty) {
```
connorhaugh, line 37: why do these have to be in the hooks, when we can just conditionally render on aspects of preexisting props? Seems just as easy to test, and way less prop passing around.

connorhaugh: Also leads to state-> render imbalances in your current setup.

src/editors/containers/TextEditor/components/SelectImageModal/messages.js
```diff
@@ -49,6 +49,16 @@ export const messages = {
...
     description: 'Gallery loading spinner screen-reader text',
   },
+  emptyGalleryLabel: {
+    id: 'authoring.texteditor.selectimagemodal.emptyGalleryLabel',
+    defaultMessage: 'No images found in your gallery. Please upload an image using the button below.',
+    description: 'Label for when image gallery is empty.',
+  },
+  emptySearchLabel: {
+    id: 'authoring.texteditor.selectimagemodal.emptySearchLabel',
+    defaultMessage: '0 search results.',
```
doctoryes, line 59: Nit: `No search results.`?

rayzhou-bit: "No search results" definitely sounds better!

src/editors/containers/TextEditor/components/SelectImageModal/hooks.js
connorhaugh: imgListHooks is called when Imghooks is called, which in theory should only be when the SelectImageModal renders or re-renders? Does that occur on uploading of an image? Just a little confused on how we can count on that logic.

src/editors/containers/TextEditor/components/SelectImageModal/Gallery.jsx
connorhaugh: Ah makes sense I didn't differntiate between those well but that makes sense. State-> render imbalance just refers to a situation where the hook isn't called again because the parent component isn't updated on a state change, so the galleryIsEmpty value is stale.

src/editors/containers/TextEditor/components/SelectImageModal/hooks.js
rayzhou-bit: I believe gallery should be re-rendering every time the displayList or list of images changes... I understand the concern on the logic not running, but this has been working when I test it.

connorhaugh (approved):
lgtm!

PR #50, opened 2022-04-07
fix: saving block title

rayzhou-bit:
I decided to change the enter-key logic to be the same as tab-key, which wasn't working as intended. It turns out that the editorRef wasn't getting passed down properly to HeaderTitle.jsx.

### DIFF
--- a/src/editors/components/EditorHeader/index.jsx
+++ b/src/editors/components/EditorHeader/index.jsx
@@ -17,11 +17,13 @@ import * as appHooks from '../../hooks';
 import HeaderTitle from './HeaderTitle';
 import messages from './messages';
 
-export const EditorHeader = ({ intl, returnUrl }) => (
+export const EditorHeader = ({ editorRef, intl, returnUrl }) => (
   <div className="editor-header">
     <ModalDialog.Header>
       <ActionRow>
-        <ModalDialog.Title><HeaderTitle /></ModalDialog.Title>
+        <ModalDialog.Title>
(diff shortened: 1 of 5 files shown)


src/editors/components/EditorHeader/index.jsx
```diff
@@ -17,11 +17,13 @@ import * as appHooks from '../../hooks';
 import HeaderTitle from './HeaderTitle';
 import messages from './messages';
 
-export const EditorHeader = ({ intl, returnUrl }) => (
+export const EditorHeader = ({ editorRef, intl, returnUrl }) => (
   <div className="editor-header">
     <ModalDialog.Header>
       <ActionRow>
-        <ModalDialog.Title><HeaderTitle /></ModalDialog.Title>
```
muselesscreator, line 24: need to update props for snapshot to include editorRef

src/editors/components/EditorHeader/index.jsx
```diff
@@ -35,7 +37,14 @@ export const EditorHeader = ({ intl, returnUrl }) => (
...
   </div>
 );
+EditorHeader.defaultProps = {
+  editorRef: null,
+};
 EditorHeader.propTypes = {
+  editorRef: PropTypes.oneOfType([
+    PropTypes.func,
+    PropTypes.shape({ current: PropTypes.any }),
+  ]),
```
muselesscreator, line 47: I would also make this isRequired

connorhaugh (approved):
Thanks for making those changes, lgtm!

PR #51, opened 2022-04-07
fix: load block title

rayzhou-bit:


src/editors/components/EditorHeader/HeaderTitle.jsx
```diff
@@ -16,7 +16,7 @@ export const HeaderTitle = ({
   intl,
   isInitialized,
   setBlockTitle,
-  typeHeader,
+  returnTitle,
```
connorhaugh, line 19: I'm a little confused what this variable does based on what I thought it was versus its name. Is this the place where a "new" title is set to?

rayzhou-bit: This was a selector that pulled the block type and the reason why we were seeing 'Text' as the title. I checked where this was used and it was only used to set up the title. Because of that, I choose to update the logic and change its name to something more apt.

connorhaugh: Ok makes sense! TY

connorhaugh:
https://user-images.githubusercontent.com/49422820/162253634-ec5eabb6-a3e3-4142-85e4-7259a9c06879.mov

connorhaugh (approved):

PR #55, opened 2022-04-07
fix: image context toolbar

connorhaugh:
Note: this removes the crop and rotate functionality from the toolbar, as it requires an image_proxy server or tinymce cloud. https://www.tiny.cloud/docs-4x/plugins/imagetools/#imagetools_proxy. We also need to create a follow up ticket to handle right click behavior. 
This is that follow up ticket


rayzhou-bit (approved):
lgtm!

PR #57, opened 2022-04-07
Tnl 9804 make small code stewardship changes to frontend lib content components

bszabo:


src/editors/components/EditorFooter/messages.js
```diff
@@ -19,9 +19,9 @@ export const messages = {
     defaultMessage: 'Save Changes and Return to Learning Context',
     description: 'Aria label for save button',
   },
-  addToCourse: {
+  saveButtonLabel: {
     id: 'authoring.editorfooter.savebutton.label',
-    defaultMessage: 'Save',
+    defaultMessage: 'Save Changes and Return to Learning Context',
```
connorhaugh, line 24: So here's a interesting thing: 
1. The design/product specifications intend that this label present "save" visually after some discussion.
2. We have an obligation to screenreader users to inform that a given piece of code leads to navigation. 

That was the motivation for the disparity between the aria-label and the button label. I'm not really comfortable with either breaking either of our obligations. Your code changes, in terms of thoroughness, is great!

connorhaugh: Also note that this change would be a `feat:` not a `refactor:`, as it changes user-facing behavior. Conventional commits are a fickle art, but I'm sure you'll come to appreciate them as I do!

bszabo: Is the Aria label only used for the screenreader? i.e., is your point that the verbose version of this label is intended for the screenreader? If so, is this message's relationship to the Aria font tangential only, and might naming of the key for this lable in terms of screenreader be more appropriate?

connorhaugh: >  your point that the verbose version of this label is intended for the screenreader? 

exactly.

> is this message's relationship to the Aria font tangential only

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label  aria = Accessible Rich Internet Applications. the font is unrelated.

> might naming of the key for this label in terms of screenreader be more appropriate?

 I believe the name authoring.editorfooter.savebutton.ariaLabel is a good name, while maybe `editorfooter.savebutton.label` could be `editorfooter.savebutton.labeltext`
 or something of that nature

bszabo: OK. Does `editorfooter.savebutton.screenreaderlabel  work?

bszabo: Your point about the need for different messages is valuable, given that's where product requirements fall. If you have a moment to educate me on what the screen reader is, and why a more verbose label is appropriate, I'd certainly appreciate it.

Given that the Aria font is unrelated, I'm confused. Why is the `author.editorfooter.savebutton.ariaLabel` a good name, over a label explicitly tied to the screenreader, as per my suggestion above?

rayzhou-bit: I believe this specific line would change the text we see on the button so it comes out like this:
image

As for naming for saveButtonAriaLabel, I believe we named it this way because the property of the Button component mentions aria-label. I'm not particularly attached to any of the naming. I would think we would want to use whatever works better for the person that does the i18n.

bszabo: I just pushed a commit that factors in this feedback, I believe. The button message is back to "Save" in places where it had been so. I changed the description in the Aria cases to reference the Screensaver, but I believe this is in keeping with what you wrote, and won't break anything. Please let me know if I got it wrong; if so, I'll gladly change it back. (PS -- I debated whether to ask before making the button message changes, but decided to let the answer surface through the PR process. Thanks for keeping me honest on this!)

bszabo:
Latest commit changes "screensaver" reference in footer Aria descriptions to "screen reader"

rayzhou-bit (approved):
lgtm!

PR #60, opened 2022-04-13
fix: enable ordered and unordered lists in text editor. Fixes TNL-9855.

kenclary:


marcotuts:
Note: I think a few other Changes got buried in the big comments, if not In this pr we may need a follow up ticket:



* Font color instead of Background color in first set of formatting rules. 
* Addition of bulleted and numbered lists
* Grouping of image + link + unlink + quotes + code formatting (unlink and quotes appear to be missing now) 
* Table formatting icon (minor nit)
* Grouping of emoji, symbols, and horizontal line (horizontal line missing) 
* fine to keep text formatting in as last tool after horizontal line (in that grouping) 
* final icon should read HTML not use the code icon

muselesscreator (approved):
lol, LGTM.  sorry for blocking you on this for so long!

PR #61, opened 2022-04-18
fix: rearrange and add several tinymce toolbar items. Addresses TNL-9874.

kenclary:


rayzhou-bit (changes requested):
Tested this using the editor gallery view.

For color selection and the ... button to see more toolbar options, the menus are displaying but not clickable. Could this have to do with the z-index issue that @connorhaugh ran into?

kenclary:
screenshot:
image

rayzhou-bit (approved):

PR #62, opened 2022-04-19
Fix  change to go back to 'fake' modal

rayzhou-bit:


src/editors/data/redux/app/selectors.test.js
```diff
@@ -67,23 +67,20 @@ describe('app selectors unit tests', () => {
     it('is memoized based on unitUrl, editorInitialized, and blockValue', () => {
       expect(selectors.isInitialized.preSelectors).toEqual([
         simpleSelectors.unitUrl,
-        simpleSelectors.editorInitialized,
         simpleSelectors.blockValue,
       ]);
     });
-    it('returns true iff unitUrl, blockValue, and editorInitialized are all truthy', () => {
```
connorhaugh, line 74: This was intentional. From math iff = "if and only if"

connorhaugh: Or has this behavior changed?

rayzhou-bit: Oh! I thought this was a typo. Let me change back.

connorhaugh: more importantly, why has this behavior changed?

connorhaugh: just confused why `editorInitialized` is being removed-- I assume its becuase editors are no longer tied to tinymce but confused why this happened in this PR in particular?

connorhaugh (approved):
Wait I get the editorintialized thing-- we need to remove it as there is no longer a callback which triggers it once tinymce loads in, and it was a bug fix in the first place. This better promotes the abstraction layers we want, so as long as it works, I'm in.

PR #65, opened 2022-04-20
feat: video skeleton hooks

muselesscreator:
Adds implementations for the duration and video source hooks, while, more importantly, providing a tested hook framework for the remainder of the collapsible widgets.
JIRA:
[TNL-9823](https://openedx.atlassian.net/browse/TNL-9823)
[TNL-9722](https://openedx.atlassian.net/browse/TNL-9772)

### DIFF
--- a/src/editors/containers/VideoEditor/components/VideoSettingsModal/components/CollapsibleFormWidget.jsx
+++ b/src/editors/containers/VideoEditor/components/VideoSettingsModal/components/CollapsibleFormWidget.jsx
@@ -3,6 +3,13 @@ import PropTypes from 'prop-types';
 
 import { Collapsible } from '@edx/paragon';
 
+/**
+ * Simple Wrapper for Form Widget components in the Video Settings modal
+ * Takes a title element and children, and produces a collapsible widget container
+ * <CollapsibleFormWidget title={<h1>My Title</h1>}>
+ *   <div>My Widget</div>
+ * </CollapsibleFormWidget>
+ */
 export const CollapsibleFormWidget = ({
   title,
   children,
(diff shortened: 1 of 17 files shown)


rayzhou-bit (approved):
lgtm

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/CollapsibleFormWidget.jsx
```diff
@@ -3,6 +3,13 @@ import PropTypes from 'prop-types';
 
 import { Collapsible } from '@edx/paragon';
 
+/**
+ * Simple Wrapper for Form Widget components in the Video Settings modal
```
connorhaugh, line 7:
```suggestion
 * Simple Wrapper for a Form Widget component in the Video Settings modal
```
nit: just was confused because I thought ALL the widgets went in this container based on the original phrasing, and was confused when I continued reading.

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/LicenseWidget.jsx
```diff
@@ -5,10 +5,18 @@ import { useDispatch } from 'react-redux';
...
+ * Collapsible Form widget controlling videe licence type and details
+ */
 export const LicenseWidget = () => {
   const dispatch = useDispatch();
-  const licenseType = hooks.widgetValue(hooks.selectorKeys.licenseType, dispatch);
-  const licenseDetails = hooks.widgetValue(hooks.selectorKeys.licenseDetails, dispatch);
+  const { licenseType, licenseDetails } = hooks.widgetValues({
+    dispatch,
+    fields: {
+      [hooks.selectorKeys.licenseType]: hooks.genericWidget,
```
connorhaugh, line 16: Curious: what caused these to be two separate widgets?

muselesscreator: Mostly because they are 2 separate data fields, with different shapes

muselesscreator: Because one is an object, and I don't want an embedded object if I can avoid it.  Makes sense for each to be an individual state field, and they are easy to control that way.

PR #66, opened 2022-04-25
feat: The X Button in the Top Right Doesn't do anything.

rayzhou-bit:
[TNL-9882 "Disabled Save button" Feedback on ImageSettings form is not sufficient] https://openedx.atlassian.net/browse/TNL-9882)

### DIFF
--- a/src/editors/containers/EditorContainer/index.test.jsx
+++ b/src/editors/containers/EditorContainer/index.test.jsx
@@ -36,16 +36,24 @@ describe('EditorContainer component', () => {
       beforeEach(() => {
         el = shallow(<EditorContainer {...props}>{testContent}</EditorContainer>);
       });
-      test('close behavior is linked to modal onClose and footer onCancel', () => {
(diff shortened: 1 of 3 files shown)


kenclary (approved):

src/editors/containers/EditorContainer/index.test.jsx
```diff
@@ -36,16 +37,23 @@ describe('EditorContainer component', () => {
       beforeEach(() => {
         el = shallow(<EditorContainer {...props}>{testContent}</EditorContainer>);
       });
-      test('close behavior is linked to modal onClose and footer onCancel', () => {
+      test('close behavior is linked to modal onClose', () => {
```
muselesscreator, line 40: Header and footer need to be mocked for the snapshots.
and you should query by finding by the element type, not by children index (unless doing both)

muselesscreator (commented):
Testing request

PR #67, opened 2022-04-26
Feat  add autocreate new xblock

connorhaugh:
This PR adds an automated script for generating the boilerplate code for adding a new xblock editor. As well as documentation for how to do so.


src/editors/containers/EditorContainer/index.scss
```diff
@@ -0,0 +1,3 @@
+.pgn__modal-layer {
```
connorhaugh, line 1: We need to remove this again, right?

www/src/Gallery.jsx
```diff
@@ -26,9 +26,10 @@ export const EditorGallery = () => {
           onChange={handleChange}
           value={blockType}
         >
-          <Form.Radio value="html">Text</Form.Radio>
-          <Form.Radio value="video">Video</Form.Radio>
-          <Form.Radio value="problem">Problem</Form.Radio>
+          { blockTypes.map((e) => (
+            <Form.Radio value={e}> {e} </Form.Radio>
+          ))}
+
```
rayzhou-bit, line 32: blockTypes is not an array here and will trigger an error.
Something like the following code should work.
```
          { Object.values(blockTypes).map((e) => (
            <Form.Radio value={e}> {e} </Form.Radio>
          ))}
```

rayzhou-bit (changes requested):
There's one bit of code that should be changed. Otherwise this looks really nice. Very simple and elegant solution to adding a new block.

rayzhou-bit (approved):
lgtm!

PR #68, opened 2022-05-02
Feat  alert users with feedback instead of disabling next/save buttons

rayzhou-bit:


src/editors/containers/TextEditor/components/ImageSettingsModal/hooks.js
```diff
@@ -7,6 +7,8 @@ import * as module from './hooks';
 export const state = {
   altText: (val) => React.useState(val),
   dimensions: (val) => React.useState(val),
+  isAltTextError: (val) => React.useState(val),
+  isAltTextValid: (val) => React.useState(val),
```
rayzhou-bit, line 11: I created two states for the error here.
isAltTextError corresponds with the alert at the top. It becomes true when save button is clicked. It becomes false when alert is dismissed.
isAltTextValid corresponds with the alt-text box and the error message near it. It becomes true or false only when the save button is clicked.

src/editors/containers/TextEditor/components/ImageSettingsModal/hooks.js
```diff
@@ -201,32 +233,43 @@ export const onCheckboxChange = (handleValue) => (e) => handleValue(e.target.che
...
+  }
+
+  if (!error) {
+    saveToEditor({
+      altText,
+      dimensions,
+      isDecorative,
+    });
+  }
+};
```
rayzhou-bit, line 269: This is less concise than can be and is left over from when I tried to write an error for dimension controls. I think leaving it in this format could be useful if additional user feedback errors are added in the future.

src/editors/containers/TextEditor/components/ErrorAlerts/FeedbackAlert.jsx
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { Alert } from '@edx/paragon';
+import { Info } from '@edx/paragon/icons';
+
```
rayzhou-bit, line 6: I created this instead of using ErrorAlert.jsx because of the Alert.Heading.

rayzhou-bit:
[TNL-9721: "Disabled Save button" Feedback on ImageSettings form is not sufficient](https://openedx.atlassian.net/browse/TNL-9721)

This change features 2 user feedback errors:
1. error prompting user to select an image on the select image gallery when pressing Next
2. error prompting user to enter alt-text or specify that the image is decorative when pressing Save

src/editors/containers/TextEditor/components/ErrorAlerts/FeedbackAlert.jsx
muselesscreator: plz just make the error title and thus heading optional in the ErrorAlert rather than repeating code.

src/editors/containers/TextEditor/components/ImageSettingsModal/AltTextControls.jsx
```diff
@@ -28,15 +31,23 @@ export const AltTextControls = ({
...
-      value={value}
       disabled={isDecorative}
-      onChange={hooks.onInputChange(setValue)}
       floatingLabel={intl.formatMessage(messages.altTextFloatingLabel)}
+      isInvalid={!errorProps.isValid}
+      onChange={hooks.onInputChange(setValue)}
+      type="input"
+      value={value}
     />
+    {errorProps.isValid
```
muselesscreator, line 41: use && for this case
`{!errorProps.isValid && (
  <Form...>
   ...
  )}`

src/editors/containers/TextEditor/components/ImageSettingsModal/hooks.js
muselesscreator: This is... confusing.  Too close to each other without enough distinction. I would rename to something like:
 - showDismissibleError
 - showSubmissionError

src/editors/containers/TextEditor/components/ImageSettingsModal/hooks.js
```diff
@@ -166,15 +178,32 @@ export const dimensionHooks = () => {
...
   const [isDecorative, setIsDecorative] = module.state.isDecorative(false);
+  const [isAltTextError, setIsAltTextError] = module.state.isAltTextError(false);
+  const [isAltTextValid, setIsAltTextValid] = module.state.isAltTextValid(true);
   return {
     value,
     setValue,
     isDecorative,
     setIsDecorative,
+    errorProps: {
+      isError: isAltTextError,
```
muselesscreator, line 200: recommended shape update:
```
error: { show, set, dismiss },
validation: { show, set, dismiss },
```

src/editors/containers/TextEditor/components/ImageSettingsModal/hooks.js
muselesscreator: If that is the goal, then you should separate out that check into a `checkFormValidation` hook (which I think would be a good idea)

src/editors/containers/TextEditor/components/ImageSettingsModal/index.jsx
```diff
@@ -38,6 +39,8 @@ export const ImageSettingsModal = ({
     dimensions: dimensions.value,
     altText: altText.value,
     isDecorative: altText.isDecorative,
+    dimensionsError: dimensions.errorProps,
```
muselesscreator, line 42: passing in but not using?

src/editors/containers/TextEditor/components/SelectImageModal/hooks.js
```diff
@@ -40,6 +41,7 @@ export const imgListHooks = ({
   const dispatch = useDispatch();
   const [images, setImages] = module.state.images({});
   const [highlighted, setHighlighted] = module.state.highlighted(null);
+  const [isSelectImageError, setIsSelectImageError] = module.state.isSelectImageError(false);
```
muselesscreator, line 44: prefer `show` to `is` for conditional rendering values.

src/editors/containers/TextEditor/components/SelectImageModal/messages.js
```diff
@@ -19,6 +19,8 @@ export const messages = {
     defaultMessage: 'Search',
     description: 'Placeholder text for search bar',
   },
+
+  // Sort Dropdown
```
muselesscreator, line 23: nice little delineation

muselesscreator (commented):
Some naming and data structure tweaks requested to improve maintainability, but this looks really good!

src/editors/containers/TextEditor/components/ImageSettingsModal/hooks.js
connorhaugh: ^ +1

src/editors/containers/TextEditor/components/ImageSettingsModal/hooks.js
rayzhou-bit: These sound better. I'll make a change for this.

src/editors/containers/TextEditor/components/ImageSettingsModal/hooks.js
rayzhou-bit: I'll code that in. Ben and I also had a conversation on when the validation should be occurring (it should be occurring onChange for alt-text field and checkbox). I'll be coding that in as well.

src/editors/containers/TextEditor/components/ErrorAlerts/FeedbackAlert.jsx
rayzhou-bit: This makes sense. Will be coding it in

src/editors/containers/TextEditor/components/ErrorAlerts/__snapshots__/ErrorAlert.test.jsx.snap
```diff
@@ -5,6 +5,7 @@ exports[`ErrorAlert component Component Snapshots snapshot:  is Null when no err
 exports[`ErrorAlert component Component Snapshots snapshot: Loads children and component when error (ErrorAlert) 1`] = `
 <Alert
   dismissible={true}
+  onClose={[Function]}
```
connorhaugh, line 8: probably want to mock this out here

connorhaugh:
Other than that nit, this looks good. Going to check it out locally and test.

connorhaugh (approved):

PR #69, opened 2022-05-03
chore: hook ADR round 1

muselesscreator:
Testability ADR.  Title up for decision


docs/decisions/000x-internal-editor-testability-decisions.md
```diff
@@ -0,0 +1,136 @@
...
+  const [myStateValue, setMyStateValue] = module.state.myStateValue('initialValue');
+};
+```
+Examples on how to use this for testing are included with the mock class in `src/testUtils`
+
+#### useCallback, useEffect
+These hooks provide behavior that calls a method based on given prerequisite behaviors.
+For this reason, we use general-purpose mocks for these hooks that return an object containing the passed callback and prerequisites for easy test access.
+
+#### Additional Considrations
```
rayzhou-bit, line 118: nit: spelling: Considerations

docs/decisions/000x-internal-editor-testability-decisions.md
```diff
@@ -0,0 +1,136 @@
+# Internal editor testability decision
+
+# Increased complexity for the sake of testability
+The internally-managed editors in this repo (as of now planned to include text, video, and problem types) follow a number of patterns that increase the complexity of parts of the code slightly, in favor of providing increased testability around their behavior.
```
rayzhou-bit, line 4: I would say that separating out the behavior in a hook files helps with readability as well.

docs/decisions/000x-internal-editor-testability-decisions.md
```diff
@@ -0,0 +1,136 @@
+# Internal editor testability decision
```
connorhaugh, line 1: nit: as this ADR contains more information about general code structure in the name of improved testability, it might be worth noting that in the title of the ADR, so those confused by the code structure in this repo know to look here.

connorhaugh: "Testability decisions" to the lay public might get skipped over because to them it means "we are shooting for 100% and we use Jest"

docs/decisions/000x-internal-editor-testability-decisions.md
```diff
@@ -0,0 +1,136 @@
...
+      expect(module.methodB(testValue)).toEqual(mockMethodA(testValue) + 3);
+    });
+  });
+});
+```
+
+## Hooks and Snapshots - Separation from components for increased viability of snapshots
+As part of the testing of these internal editors, we are relying on snapshot testing to ensure stability of the display of the components themselves.  This can be a fragile solution in certain situations where components are too large or complex to adequately snapshot and verify.
+
+For this purpose, we have opted for a general pattern of separating all of the behavior of components withing these editors into separate `hooks` files.
```
connorhaugh, line 54: also sometimes not just files, but static objects defined in the same file as the component, when small.

connorhaugh (commented):
Just a couple of small things-- this is really great to capture "goals" we have had as concrete patterns. We should also double check we aren't breaking any StrictDict imports somewhere which imports it directly from './editors/utils'

docs/decisions/000x-internal-editor-testability-decisions.md
```diff
@@ -0,0 +1,136 @@
...
+The internally-managed editors in this repo (as of now planned to include text, video, and problem types) follow a number of patterns that increase the complexity of parts of the code slightly, in favor of providing increased testability around their behavior.
+
+## Note - Strict Dictionaries
+Javacript is generally fairly lackadaisical with regards to dictionary access of missing/invalid keys.  This is fine and expected in many cases, but also prevents us from using dictionary access on something like a key store to ensure we are calling something that actually exists.
+
+For this purpose, there are a pair of utilities in this repo called `StrictDict` and `keyStore`.
+
+`StrictDict` takes an object and returns a version that will complain (throw an error) if called with an invalid key.
+
+`keyStore` takes an object and returns a StrictDict of just the keys of that object.  (this is useful particularly for mocking and spying on specific methods and fields)
```
farhaanbukhsh, line 13: (/) We should add this in the comments where we have declared these utilities.

docs/decisions/000x-internal-editor-testability-decisions.md
```diff
@@ -0,0 +1,136 @@
...
+
+These hook files contain methods that utilize both `react` and `react-redux` hooks, along with arguments passed directly into the component, in order to generate the resulting desired behaviors.
+
+From there, components are tested by mocking out the behavior of these hooks to return verifyable data in the snapshots.
+
+As part of this separation, there are a number of additional patterns that are followed 
+
+### Snapshot considerations
+#### Callbacks
+Any callback that is included in render in a component should be separated such that is is either passed in as a prop or derived from a hook, and should be mocked with a `mockName` using jest, to ensure that they are uniquely identifyable in the snapshots.
```
farhaanbukhsh, line 64:
```suggestion
Any callback that is included in render in a component should be separated such that it is either passed in as a prop or derived from a hook, and should be mocked with a `mockName` using jest, to ensure that they are uniquely identifyable in the snapshots.
```

farhaanbukhsh:
This is great work ⭐, adding more examples, links to the PRs, and commit will make the document gold.

docs/decisions/000x-internal-editor-testability-decisions.md
jesperhodge, line 4: I am not sure that idea makes sense. What usually happens is that hard-to-test components are relatively complex, and reducing that complexity makes them easy to test. If you need to increase code complexity to test better, it is more likely a sign that the code is already too complex (and then you make matters worse.) I know testing is hard but the better way is to decouple your code enough that it's testable, develop better testing tools, and also testing the right things (I recommend starting to use react-testing-library to make your life easier here.)

docs/decisions/000x-internal-editor-testability-decisions.md
```diff
@@ -0,0 +1,136 @@
...
+#### Top-level mocked imports
+We have mocked out all paragon components and icons being used in the repo, as well as a number of other common methods, hooks, and components in our module's `setupTests` file, which will ensure that those components show up reasonably in snapshots.
+
+### Hook Considerations
+#### useState and mockUseState
+React's useState hook is a very powerful alternative to class components, but is also somewhat problematic to test, as it returns different values based on where it is called in a hook, as well as based on previous runs of that hook.
+
+To resolve this, we are using a custom test utility to mock a hook modules state values for easy testing.
+
+This requires a particular structure to hook modules that use the useState, for the sake of enabling the mock process (which is documented with the utility).
```
jesperhodge, line 100: This doesn't make sense, because we are making the code much much more complex and hard to understand. This violates the clean code principles but also is dangerous as you don't understand the state logic of the component easily and it can produce bugs. Changing implementations to make them easier testable is really not a good idea, instead, you can improve the way you test. It is absolutely possible to mock React's useState without changing the code, it is just very hard. Which is okay, because you are not supposed to mock it in general. When you need to mock it, it may be a sign that your test does not test the right thing. Unit testing means not breaking into the implementation of a component but testing it from outside, more like a user uses it. React recommends trying to test a component much like a user would - e.g. you would simulate a click on a dom element rather than overwriting the state, which is not very helpful. To make your life easier with this, I recommend react-testing-library in many cases over enzyme.
If you have trouble testing complex state logic, consider: Redux in part is designed to make this easy and make state testable. So maybe use a reducer and test that separately. Also, React has its own useReducer hooks if you don't want to interact with global state for some reason.

jesperhodge (changes requested):

docs/decisions/000x-internal-editor-testability-decisions.md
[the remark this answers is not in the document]
connorhaugh: What's the effective utility of having two separate files for "non-hook helper functions" and "hooks"? Can we just name these files "logic.js" or something like that instead of Hooks (which I agree is a term we have overloaded)?

docs/decisions/000x-internal-editor-testability-decisions.md
```diff
@@ -0,0 +1,136 @@
...
+`StrictDict` takes an object and returns a version that will complain (throw an error) if called with an invalid key.
+
+`keyStore` takes an object and returns a StrictDict of just the keys of that object.  (this is useful particularly for mocking and spying on specific methods and fields)
+
+## Note - Self imports
+Javascript webpack imports can be problematic around the specific issue of attempting to mock a single method being used in another method in the same file.  
+
+Problem: File A includes methodA and methodB (which calls methodA).  We want to be able to test methodA and then test methodB *without* having to re-test methodA as part of that test.  We want to be able to mock methodA *only* while we are testing methodB.
+
+Solution: Self-imports.  By importing the module into itself (which webpack handles nicely, don't worry), we provide tests the ability to spy on and mock individual methods from that module separately on a per-test basis.
```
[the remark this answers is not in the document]
connorhaugh: > very dangerous and bug-prone
I'm sure it can be bug prone, how is it dangerous. As someone who has only scratched the surface of javascript I am curious

docs/decisions/000x-internal-editor-testability-decisions.md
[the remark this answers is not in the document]
jesperhodge: Sure, that works

docs/decisions/000x-internal-editor-testability-decisions.md
jesperhodge: That's a very good point, and I would have to figure it out a bit for myself - in previous projects I tested hooks as part of component and that's it. But there are a lot of good resources for testing custom hooks, I think even whole libraries for this purpose. A bit of research provided this very good post about it by Kent C.Dodds, who is a well-known authority in the React community, and I think his approach is very valid: https://kentcdodds.com/blog/how-to-test-custom-react-hooks

jesperhodge: I think testing hooks without using a React component is a bit beside the point because you are actually trying to test the component lifecycle, and that would be very far removed from reality. Any non-lifecycle related logic you can factor out of those hooks and test very easily. Personally so far I used react-testing-library (https://testing-library.com/docs/react-testing-library/intro/) for all of those cases.

jesperhodge: By the way the `renderHook` method from @testing-library/react Dodds uses for testing custom hooks is the same thing as react-testing-library

jesperhodge: a point I overlooked here in Dodd's article is "I'm not talking about the one-off custom hook you pull out just to make your component body smaller and organize your code (those should be covered by your component tests)" - I agree with that

docs/decisions/000x-internal-editor-testability-decisions.md
jesperhodge: "decision on structuring code for testability" maybe or something similar

jesperhodge:
PR replaced with #201 and closed

docs/decisions/000x-internal-editor-testability-decisions.md
```diff
@@ -0,0 +1,136 @@
+# Internal editor testability decision
+
+# Increased complexity for the sake of testability
```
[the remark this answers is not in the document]
kenclary: I think what we're trying to do is create/maintain separation of logic and rendering; basically, recreating a view/controller split in react. This (in theory) simplifies testing, by allowing the dom rendering to be tested in a static fashion, while actual javascript logic can be tested in a straightforward way. I think there's value in this, even if we don't maintain 100% separation.

docs/decisions/000x-internal-editor-testability-decisions.md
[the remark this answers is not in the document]
kenclary: I agree that self-imports is needlessly complex and makes things worse.

Furthermore, I think that using it to mock the functions we are testing when testing higher-level code is exactly what's making our tests horribly complicated and difficult to read, maintain, or update.

PR #77, opened 2022-05-16
feat: allow all html tags

connorhaugh:


kenclary (approved):
Fix lint, but otherwise looks good.

PR #80, opened 2022-05-21
feat: add simple code block and blockquote buttons

connorhaugh:
adresses
https://2u-internal.atlassian.net/browse/TNL-9836
and 
https://2u-internal.atlassian.net/browse/TNL-9936

### DIFF
--- a/src/editors/containers/TextEditor/__snapshots__/index.test.jsx.snap
+++ b/src/editors/containers/TextEditor/__snapshots__/index.test.jsx.snap
@@ -56,6 +56,7 @@ exports[`TextEditor snapshots block failed to load, Toast is shown 1`] = `
           "openModal": [MockFunction modal.openModal],
           "setEditorRef": [MockFunction hooks.prepareEditorRef.setEditorRef],
           "setSelection": [MockFunction hooks.selectedImage.setSelection],
+          "studioEndpointUrl": "sOmEoThERvaLue.cOm",
         }
       }
     />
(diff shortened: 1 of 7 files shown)


rayzhou-bit (changes requested):
I tested this out using the www gallery.

- Add image button works great!
- I couldn't get the edit image button to appear.
- Blockquote looks great!
- Code block button isn't highlighted the same way blockquote is when the cursor is in code block content. This behavior is the same as the old editor. So my question is should this behavior remain?

kenclary (approved):
nice.

connorhaugh:
@rayzhou-bit 
> I tested this out using the www gallery.
Just to note, I have run on devstack as well and it is working.

> * I couldn't get the edit image button to appear.
 I have added that functionality back in see the attached screenshot
Screen Shot 2022-05-24 at 9 37 15 AM

> * Code block button isn't highlighted the same way blockquote is when the cursor is in code block content. This behavior is the same as the old editor. So my question is should this behavior remain?
Yeah this is a symptom of <code> being an inline html element, instead of a block element. I don't want to rewrite the whole Tinymce html formatter we got for free, and making a custom block format will make things worse.

connorhaugh:
@rayzhou-bit  headed out soon but would love your thoughts on my changes tysm.

rayzhou-bit:
I tested the changes and it looks good to me!

rayzhou-bit (approved):

PR #81, opened 2022-05-25
fix: allow style tags inside html body, to match old editor behavior. TNL-9943.

kenclary:


connorhaugh (approved):
Tested, it looks good. sorry for the hullabaloo

PR #88, opened 2022-07-01
feat: www gallery raw selection

rayzhou-bit:
https://2u-internal.atlassian.net/browse/TNL-9998

The www gallery is a tool for developers contained in the `/www` folder of this repo. Running `npm start` through this folder starts a local instance that is used for quicker turn around in development (faster than make down/up on devstack). This code change aims to add a visual/raw selection such that the raw editor can be viewed in this gallery view.

There are 2 variables in determining whether or not we see the raw editor: `isRaw` and `mockRaw`. `isRaw` is determined by data we retrieve from the api. `mockRaw` is determined by a selection made in the gallery view (and is implemented in this change). One thing I am worried about for this change is if passing down the `mockRaw` prop from the gallery view and down through the editor components would be intrusive to the editor code.

Another way I've thought about doing this is changing the `mockApi` call depending on whether or not raw is selected. However, this could mean some messy state management.

### DIFF
--- a/src/editors/hooks.js
+++ b/src/editors/hooks.js
@@ -8,7 +8,7 @@ import * as module from './hooks';
 
 export const initializeApp = ({ dispatch, data }) => useEffect(
   () => dispatch(thunkActions.app.initialize(data)),
-  [],
+  [data],
 );
 
 export const navigateTo = (destination) => {
--- a/www/src/Gallery.jsx
+++ b/www/src/Gallery.jsx
@@ -37,7 +37,13 @@ export const EditorGallery = () => {
             ))}
           </Form.RadioSet>
         </Form.Group>
-        <Form.Group style={{ display: blockType === 'html' ? 'block' : 'none' }}>
+        <Form.Group
+          style={{
+            display: blockType === 'html' || blockType === 'problem'
+              ? 'block'
+              : 'none',
+          }}
+        >
           <Form.RadioSet
             name="raw"
             onChange={handleRawChange}
--- a/www/src/Gallery.jsx
+++ b/www/src/Gallery.jsx
@@ -10,6 +10,8 @@ import { mockBlockIdByType } from '@edx/frontend-lib-content-components/editors/
 
 export const EditorGallery = () => {
   const [blockType, setBlockType] = React.useState('html');
+  const [mockRaw, setMockRaw] = React.useState(false);
+
   const blockIds = Object.keys(blockTypes).reduce((obj, blockTypeKey) => {
     const type = blockTypes[blockTypeKey];
     return { ...obj, [type]: mockBlockIdByType(type) };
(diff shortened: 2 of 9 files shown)


rayzhou-bit:
I have notice that `mockRaw` bleeds into the system and that's my main concern for the code I've written. I originally looked into changing `fetchStudioView` in the mockApi like what Connor mentioned. I could see making a change either there or in `/editors/data/redux/thunkActions/requests.js` to look to a test redux store for the config value.

Another slightly related thought I've had is moving mockApi out into the `www` folder, because it is essentially test code. The downside for this is that it's much clearer to keep it next to api.js.

www/src/Gallery.jsx
```diff
@@ -10,28 +10,50 @@ import { mockBlockIdByType } from '@edx/frontend-lib-content-components/editors/
 
 export const EditorGallery = () => {
   const [blockType, setBlockType] = React.useState('html');
+  const [mockRaw, setMockRaw] = React.useState(false);
+
   const blockIds = Object.keys(blockTypes).reduce((obj, blockTypeKey) => {
     const type = blockTypes[blockTypeKey];
     return { ...obj, [type]: mockBlockIdByType(type) };
   }, {});
```
connorhaugh, line 18:
```suggestion
  }, {});
  
  // and add a blockid here that contians whay you want and is of the blocktype HTML
```

connorhaugh: you might need to make one more level of abstraction here (keep a list of all the views we want to present, not just block types, but you catch my drift)

src/editors/hooks.js
```diff
@@ -8,7 +8,7 @@ import * as module from './hooks';
 
 export const initializeApp = ({ dispatch, data }) => useEffect(
   () => dispatch(thunkActions.app.initialize(data)),
-  [],
+  [data],
```
rayzhou-bit, line 11: I updated editor initialization to have dependency on data which would be `blockId, blockType, learningContextId,
 lmsEndpointUrl, studioEndpointUrl` so useEffect will run when `blockId` updates.

Outside of testing, I don't see these fields changing... so I think this change to the editor code would be fine.

connorhaugh: Yeah, this dependacy array is such that it is all values which are static, so I don;t think it would be too complex a behavior change under the hood. Might be worthwhile to smoke test in on devstack for sure.

connorhaugh:
Looks like a much cleaner code change: I think Bernie will like this! If I could get a video of it working, that would be awesome, as well as some confirmation things work on devstack as expected. Maybe try putting a log in on the UseEffect and run in devstack to see if you trigger unexpected changes to the dependency array or any slowdown.

rayzhou-bit:
image
I did a console log of `data` passed into `useEffect` in devstack. These are not values we would change when we navigate onto an editor page, not for text or any future editors I think. Only one instance of it triggers as expected so this shouldn't affect performance.

bszabo:
After my deep dive on text editor internals, I'm in a much better position to articulate what I was looking for. I realize this input is probably too late, but at least I can clarify what my intuition and prior experience was leading me to.

Today, `EditorPage` is the react component that's the entry point into `frontend-lib-content-components`. It's the gateway irrespective of whether `fronted-app-course-authoring` is the client or whether the `www` test folder is the client.

`EditorPage` is initialized with `lmsEndpointUrl` and `studioEndpointUrl` props.  `EditorPage` passes these props values to `Editor`, which then passes them to the `src/editors/hooks.js` `initializeApp` hook. Here, they are added to the `app` redux module by a reducer and saved as state.

The story continues with `src/editors/data/services/cms/api.js`. This file is actually misnamed, and would more accurately be named `cmsClient` as its role is to marshal requests into CMS services. Specifically, it issues GET requests into studio to fetch blocks, units, views, and images, and it issues POST requests to save blocks and images.

When viewed in this light, it becomes clear that cmsClient is something that should be injected into `EditorPage`. As a client into CMS services, it is providing access to a peer dependency for the `frontend-lib-content-components` client, and is only very weakly coupled with `frontend-lib-content-components` capabilities. In truth, all that `frontend-lib-content-components` needs is an externally supplied client object, and it's 100% OK for the implementation behind that object to be opaque to `frontend-lib-content-components`.

From an implemenatation perspective, the following changes would be required:

1. Move the `services/cms` folder out of the `frontend-lib-content-components` and give it a home in `frontend-app-course-authoring`
2. Move the mockApi.js file to the `www` folder
3. Rename `api.js` to `cmsClient`
4. Take any business logic presently in `api.js` and hoist it out. The file (now `cmsClient`) should do nothing beyond blindly assembling requests from api call arguments and forwarding these to studio.
4. Stop propagating `lmsEndpointUrl` and studioEndpointUrl` props values; replace them with a cmsClient object
5. Add logic someplace in `frontend-lib-content-components` that verifies that the cmsClient object injected into it has definitions for all required methods

rayzhou-bit:
@bszabo What you're suggesting makes a lot of sense to me from the code perspective, but I am seeing a different problem in terms of moving `services/cms`. My understanding of `frontend-lib-content-components` as an MFE was that it was intended to bring code out from the monolith. Moving code into `frontend-app-course-authoring` does not seem like a direction we should move in.

bszabo (approved):

PR #92, opened 2022-07-29
fix: added underline option to toolbar. TNL-10003

KristinAoki:
## Description
This PR adds the missing text underline button and functionality to the Text Editor toolbar. To add the functionality, a new parameter was added to the tinyMCE configuration. No new tests required since toolbar tests are already present and cover the new button.
Jira Ticket: [TNL-10003](https://2u-internal.atlassian.net/browse/TNL-10003?atlOrigin=eyJpIjoiZGU2MTNmZDNhNzdkNDZlOWJjZDAzMGU3MWZmYzY1ZjQiLCJwIjoiaiJ9)

### DIFF
--- a/src/editors/containers/TextEditor/pluginConfig.js
+++ b/src/editors/containers/TextEditor/pluginConfig.js
@@ -21,7 +21,7 @@ export default StrictDict({
   toolbar: mapToolbars([
     [buttons.undo, buttons.redo],
     [buttons.formatSelect],
-    [buttons.bold, buttons.italic, buttons.foreColor, buttons.backColor],
+    [buttons.bold, buttons.italic, buttons.underline, buttons.foreColor, buttons.backColor],
     [
       buttons.align.left,
       buttons.align.center,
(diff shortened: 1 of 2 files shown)


kenclary:
Looks like the only problem is the commit message. You probably want something like "fix: added underline option to toolbar. TNL-10003" to pass lint.

bszabo:
I have a couple of generic comments. With regards to the PR description, you've provided a link to the underlying JIRA ticket and that's both essential and great. Thank you. Could you please either paraphrase or cut-and-pasted the essential points from that ticket here, for direct context to the PR? The acceptance criteria are especially important. In this case, they read "The Underline button is in the toolbar, and functions as expected." 

It's unclear from the code changes that the button functions as expected. Specifically, the button should now appear on the text editor, but it should also function. A brief discussion here as to how underlining relates to native tinyMCE capabilities would be in order. Also, if you did not add unit tests for this feature, an explanation is in order (i.e., the tests were already in the package and were confirmed to pass; alternatively, this is a highly visual function that doesn't lend itself to unit testing, etc.)

bszabo (approved):

PR #94, opened 2022-08-03
feat: Add functionality to resize image based on percentage

KristinAoki:
This PR adds the ability to resize an image with percentages. I changed the height and width input fields from number fields to string fields. With this change, it still prevents non-numeric inputs except if there is a percent sign present. No new tests were required, but tests and snapshots were updated to reflect the input type change and state variable name changes.
Jira Ticket: [TNL-9972](https://2u-internal.atlassian.net/browse/TNL-9972)


src/editors/containers/TextEditor/components/ImageSettingsModal/DimensionControls.jsx
```diff
@@ -42,18 +42,14 @@ export const DimensionControls = ({
     <div className="mt-4.5">
       <Form.Control
         className="dimension-input"
-        type="number"
```
connorhaugh, line 45: Did you look into the feasibility of making the accepted pattern integers and the percent symbol and preventing default otherwise? the freedom to might make people add 'px', etc.

connorhaugh:
Looks like you might have accidentally uploaded a new package-lock.json in your PR. It might be worthwhile not to upload that as you didn't change any of the dependancies.

src/editors/containers/TextEditor/components/ImageSettingsModal/hooks.js
```diff
@@ -157,8 +158,20 @@ export const dimensionHooks = () => {
     lock,
     unlock,
     value: local,
-    setHeight: (height) => setLocal({ ...local, height: parseInt(height, 10) }),
-    setWidth: (width) => setLocal({ ...local, width: parseInt(width, 10) }),
+    setHeight: (height) => {
```
connorhaugh, line 161: I think maybe we should set some kind of[ feedback](https://paragon-openedx.netlify.app/components/form/form-control-feedback/) or error if the user puts in things other that numbers and the percent sign in because right now an acceptable input is "connor%"

connorhaugh (changes requested):

bszabo:
I really like that you carried over my guidelines for writing a good PR description from the last PR! Thank you.

src/editors/containers/TextEditor/components/ImageSettingsModal/hooks.js
connorhaugh: Although I'm not super keen on not providing any feedback when the user enters in Connor%, we were told to not make any UI changes, so I guess we will live with that.

connorhaugh:
Thanks for taking a look at my suggestions and incorporating them!  I will also suggest you take a look at this, and then I can thumb :) 

> Looks like you might have accidentally uploaded a new package-lock.json in your PR. It might be worthwhile not to upload that as you didn't change any of the dependancies.

connorhaugh:
Ah, I should have been more clear, https://stackoverflow.com/questions/39459467/remove-a-modified-file-from-pull-request, we don't want to remove the file from master, we want to remove your changes to it.

connorhaugh (approved):
Looks great sorry!

PR #96, opened 2022-08-09
feat: add altText and isDecorative persistence for image editing

KristinAoki:
This PR fixes the persistence issues with altText and decorative details when editing an image. To fix the persistence, a call to update the values when loading the image was missing. Previously there was only a call to update the image dimension fields. 
Jira Ticket: [TNL-9953](https://2u-internal.atlassian.net/browse/TNL-9953)


kenclary (approved):
nice to see this fixed.

PR #97, opened 2022-08-11
feat: update editor text to match studio styling

KristinAoki:
This PR addresses the font discrepancy in studio and the new editor. The editor was set to use the default styling, so the font that is used in studio was never set. The editor is now set to use the same font as studio with the `content_style` attribute.
Jira Ticket [TNL-10057](https://2u-internal.atlassian.net/browse/TNL-10057)


kenclary (approved):
more control is good

PR #102, opened 2022-08-18
feat: run onImgLoad even on broken images

rayzhou-bit:
https://2u-internal.atlassian.net/browse/TNL-9973

When editing existing images, the UI fails to generate properly on images with broken links. This is because we are running the load logic on the image's onLoad event. This causes issues down the line with saving the dimensions to a broken image.

### DIFF
--- a/src/editors/containers/TextEditor/components/ImageSettingsModal/__snapshots__/index.test.jsx.snap
+++ b/src/editors/containers/TextEditor/components/ImageSettingsModal/__snapshots__/index.test.jsx.snap
@@ -73,7 +73,9 @@ exports[`ImageSettingsModal render snapshot 1`] = `
           Object {
             "hooks.dimensions.onImgLoad.callback": Object {
               "selection": Object {
-                "selected": "image data",
+                "altText": "AlTTExt",
+                "externalUrl": "ExtERNALurL",
+                "url": "UrL",
               },
             },
           }
@@ -82,11 +84,14 @@ exports[`ImageSettingsModal render snapshot 1`] = `
           Object {
             "hooks.dimensions.onImgLoad.callback": Object {
               "selection": Object {
-                "selected": "image data",
+                "altText": "AlTTExt",
+                "externalUrl": "ExtERNALurL",
+                "url": "UrL",
               },
             },
           }
         }
+        src="ExtERNALurL"
       />
     </div>
     <hr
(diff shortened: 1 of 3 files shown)


src/editors/containers/TextEditor/components/ImageSettingsModal/__snapshots__/index.test.jsx.snap
```diff
@@ -95,6 +104,11 @@ exports[`ImageSettingsModal render snapshot 1`] = `
                   "selected": "image data",
                 },
               ],
+              Array [
```
connorhaugh, line 107: How did these two sections get added to the snapshot if no other were changed?

src/editors/containers/TextEditor/components/ImageSettingsModal/__snapshots__/index.test.jsx.snap
```diff
@@ -107,6 +121,16 @@ exports[`ImageSettingsModal render snapshot 1`] = `
                   },
                 },
               },
+              Object {
```
connorhaugh, line 124: Nothing should have been added to the dimensioncontrols object right?

src/editors/containers/TextEditor/components/ImageSettingsModal/__snapshots__/index.test.jsx.snap
```diff
@@ -69,6 +69,15 @@ exports[`ImageSettingsModal render snapshot 1`] = `
     >
       <Image
         className="img-settings-thumbnail"
+        onError={
```
connorhaugh, line 72: This part makes sense to me

src/editors/containers/TextEditor/components/ImageSettingsModal/__snapshots__/index.test.jsx.snap
rayzhou-bit: Even if something changed about onImgLoad, DimensionControls doesn't read that value anyways.

connorhaugh: Yeah v confused-- we might have to did into what the test is doing here.

connorhaugh (approved):
Ok I think these snapshots map now, but I'm confused as to how this happened?

rayzhou-bit:
This still doesn't make sense to me... all I can tell is that the number of entries in the "calls" and "results" array in the DimensionControls snapshot corresponds to the number of times dimensions.onImgLoad(selection) is called.

When I add more dimensionsImgLoad entries or change the arg I'm tossing in, the "calls" array will grow and change in kind. It's the same for when I remove entries. I can't figure out how to mock dimensions.onImgLoad to not show the "calls" and "results" array either.

At this point, it's definitely not worth it to sink more time into it. What's being passed into DimensionControls is the function without any args and DimensionControls.js itself is not even looking at onImgLoad.

PR #103, opened 2022-08-22
feat: change static url to asset url in editor

KristinAoki:
This PR changes `/static/` links to `/assets/` links. This fixes the broken links in the editor view. The links are automatically replaced when the source code is saved and the user is navigated back to the main body of the editor. Updated tests to reflect the new way that `fetchImages` is called and how images are passed between components.

Linked PR:  edx-platform [PR #30886](https://github.com/openedx/edx-platform/pull/30886)

Jira Ticket: [TNL-10077](https://2u-internal.atlassian.net/browse/TNL-10077)

### DIFF
--- a/src/editors/containers/TextEditor/components/SelectImageModal/hooks.js
+++ b/src/editors/containers/TextEditor/components/SelectImageModal/hooks.js
@@ -29,12 +29,11 @@ export const filteredList = ({ searchString, imageList }) => (
   imageList.filter(({ displayName }) => displayName.toLowerCase().includes(searchString.toLowerCase()))
 );
 
-export const displayList = ({ sortBy, searchString, images }) => {
-  return (
+export const displayList = ({ sortBy, searchString, images }) => (
   module.filteredList({
     searchString,
     imageList: Object.values(images),
-  }).sort(sortFunctions[sortBy in sortKeys ? sortKeys[sortBy] : sortKeys.dateNewest]))};
+  }).sort(sortFunctions[sortBy in sortKeys ? sortKeys[sortBy] : sortKeys.dateNewest]));
 
 export const imgListHooks = ({ searchSortProps, setSelection, images }) => {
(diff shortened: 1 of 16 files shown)


src/editors/data/redux/thunkActions/app.js
```diff
@@ -35,6 +41,7 @@ export const initialize = (data) => (dispatch) => {
   dispatch(module.fetchBlock());
   dispatch(module.fetchUnit());
   dispatch(module.fetchStudioView());
+  dispatch(module.fetchImages());
```
rayzhou-bit, line 44: This is great! Makes much more sense to fetch at start rather than when we open up gallery.

src/editors/containers/TextEditor/components/SelectImageModal/hooks.js
```diff
@@ -36,9 +35,7 @@ export const displayList = ({ sortBy, searchString, images }) => (
     imageList: Object.values(images),
   }).sort(sortFunctions[sortBy in sortKeys ? sortKeys[sortBy] : sortKeys.dateNewest]));
 
-export const imgListHooks = ({ searchSortProps, setSelection }) => {
-  const dispatch = useDispatch();
-  const [images, setImages] = module.state.images({});
+export const imgListHooks = ({ searchSortProps, setSelection, images }) => {
```
rayzhou-bit, line 38: When I try running this in www gallery, I get the error `Cannot convert undefined or null to object`. I'm guessing this has to do with images being null by default.

KristinAoki: yes, and gallery doesn't use any api calls so it won't be able to pull anything. Should I make it so that it defaults to an empty object?

rayzhou-bit: Yes, I think that should work.

PR #104, opened 2022-08-26
feat: add codeMirror lineWrapping extension

KristinAoki:
This PR address the horizontal scroll when editing raw html. Previously the user would manually have to enter for new lines or the line would stretch on forever. This makes it difficult to read the html. Now the text will automatically wrap the line when it hits the edge of the container.

JIRA Ticket: [TNL-10099](https://2u-internal.atlassian.net/browse/TNL-10099)


kenclary (approved):
awesome

PR #105, opened 2022-08-29
feat: remove the ability to use image tools for content libraries

KristinAoki:
This PR removes the functionality of images in the text editor for content libraries, but retains the functionality for course blocks. Updated test to include the check for if the editor is editing a library block.

JIRA Ticket: [TNL-9958](https://2u-internal.atlassian.net/browse/TNL-9958)

### DIFF
--- a/src/editors/data/redux/app/selectors.js
+++ b/src/editors/data/redux/app/selectors.js
@@ -83,12 +83,12 @@ export const isRaw = createSelector(
 );
 
 export const isLibrary = createSelector(
-  [module.simpleSelectors.blockId],
-  (blockId) => {
-    if (!blockId) {
+  [module.simpleSelectors.learningContextId],
+  (learningContextId) => {
+    if (!learningContextId) {
       return null;
     }
-    if (blockId.startsWith('lib')) {
+    if (learningContextId && learningContextId.startsWith('library-v1'))  {
       return true;
     }
     return false;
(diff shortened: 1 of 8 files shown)


src/editors/data/redux/app/selectors.js
```diff
@@ -82,11 +82,25 @@ export const isRaw = createSelector(
   },
 );
 
+export const isLibrary = createSelector(
+  [module.simpleSelectors.blockId],
```
connorhaugh, line 86: Why are we using blockID and not learningcontextID here? Just from https://github.com/edx/frontend-lib-content-components/commit/cc4e19cd2abee9249873f5c0507a8b2791dbcdf1 we used learning contextID

connorhaugh: Also, we should probably have a central util which determines if something is a library so having these two paths match in what they check (`if (learningContextId && learningContextId.includes('library-v1')) {` versus ` if (blockId.startsWith('lib')) {` )Should be the same kind of check.

KristinAoki: I used `blockId` because I knew that was specific for a block. I didn't notice the `learningContextId`.  But I will update it to match.

connorhaugh (approved):
Looks Good to me, thanks for aligning these

PR #106, opened 2022-08-29
feat: add tinyMCE functionality for drag resizing of images

KristinAoki:
This PR adds the missing CSS from tinyMCE that allows for image resizing by dragging. Solution based on [tinymce-react Issue 56](https://github.com/tinymce/tinymce-react/issues/56).

JIRA ticket: [TNL-10084](https://2u-internal.atlassian.net/browse/TNL-10084)

### DIFF
--- a/src/editors/data/constants/tinyMCEStyles.js
+++ b/src/editors/data/constants/tinyMCEStyles.js
@@ -1,4 +1,101 @@
 export default `@import url("https://fonts.googleapis.com/css2?family=Open+Sans:ital,wght@0,300;0,400;0,600;0,700;1,300;1,400;1,600;1,700&display=swap");
+
+.mce-content-body *[contentEditable=false] {
+  cursor: default;
+}
+.mce-content-body *[contentEditable=true] {
+  cursor: text;
+}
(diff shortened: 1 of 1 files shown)


connorhaugh:
Why was the CSS missing before? Do we have to regularly update the CSS because we are not pulling from tinyMCE cloud? How might this code change when we try to upgrade tinymce for some particular reason?

KristinAoki:
The CSS was missing before because we are not using the API/cloud version. The CSS has to given to tinyMCE in a literal string. I decided to only keep the CSS that allows for resizing. However, there is another way that I could do it so that is always pulls the current styling in the node module and converts it to usable string. There is a [GitHub Issue](https://github.com/tinymce/tinymce-react/issues/56) that discusses the problem we are experiencing. The second option requires me to install another library, which I do not know much about.

connorhaugh:
If the `raw-loader` doesn't work out of the box with webpack (which we install through I think frontend-build already) then I think this is an acceptable solution. Can you confirm that it requires some other install other than tinymce and webpack?

connorhaugh (approved):

PR #107, opened 2022-08-29
Feat  shared widget componentries and layout

rayzhou-bit:
https://2u-internal.atlassian.net/browse/TNL-9770

This change implements the shared componentry for widgets on the Video Editor. This includes implementation of the advanced Collapsible component from paragon, a framework for error states and validating the various widgets when attempting to save.

Testing:
1. Run `npm start` from the /www folder.
2. Watch the Transcript widget for the various changes as this is where I wrote some mock props.
3. Collapsing the Transcript widget should show a subtitle of the transcript name.
4. When pressing the Save button, the fields for each widget is validated. This is currently mocked to return an error in the Transcript widget, which displays a generic error message at the top.
5. Going to the Transcript widget, it can no longer be closed as it should remain open to display errors shown in its fields (which has yet to be implemented).

Error data structure is intended to look like the following:
```
errror: {
  duration: {
    [fieldName]: "error text",
    [fieldName]: "error text",
  },
  handout: {
    [fieldName]: "error text",
    [fieldName]: "error text",
  },
  ...
}
```


src/editors/containers/VideoEditor/hooks.js
```diff
@@ -0,0 +1,80 @@
+import { useState } from 'react';
```
rayzhou-bit, line 1: I chose to place the logic at this level because of validateEntry. This is what checks for errors when we hit the Save button. This gets passed to the EditorContainer. The errors themselves can get passed down through the components from this level.

src/editors/containers/VideoEditor/components/VideoEditorModal.jsx
```diff
@@ -16,21 +16,31 @@ export const hooks = {
...
   module.hooks.initialize(dispatch);
   return (
-    <VideoSettingsModal {...{ isOpen, close }} />
+    <VideoSettingsModal {...{ close, error, isOpen }} />
   );
   // TODO: add logic to show SelectVideoModal if no selection
 };
 
 VideoEditorModal.defaultProps = {
+  error: {
```
connorhaugh, line 32: Have you thought about other data structures for use to represent errors?

connorhaugh: I'm thinking about how we might have multiple failed validations for a single widget?

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/CollapsibleFormWidget.jsx
```diff
@@ -11,16 +12,52 @@ import { Collapsible } from '@edx/paragon';
...
+      style={{ justifyContent: 'unset' }}
+    >
+      <Collapsible.Visible whenClosed>
+        <div className="d-flex flex-column flex-grow-1">
+          <div className="d-flex flex-grow-1 w-75">{title}</div>
+          {subtitle}
+        </div>
+        <div className="d-flex flex-row align-self-start">
+          {isError && <Icon className="alert-icon" src={Info} />}
+          <IconButton alt="TODOformatted" src={ExpandMore} iconAs={Icon} variant="dark" />
```
connorhaugh, line 36: alt text here is a todo

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/hooks.js
```diff
@@ -205,15 +205,15 @@ export const arrayWidget = ({ dispatch, key }) => {
   const widget = module.valueHooks({ dispatch, key });
   return {
     ...widget,
-    onChange: handleIndexTransformEvent({
+    onBlur: handleIndexTransformEvent({
```
connorhaugh, line 208: this is an alphabetical change? want to be sure I'm reading it right

src/editors/containers/VideoEditor/components/VideoEditorModal.jsx
rayzhou-bit: I should have mentioned but I imagined the data structure to look like the below. I'm open to suggestions on other ways to do this. (I'll update the top level comment to describe this.)
```
errror: {
  duration: {
    [fieldName]: "error text",
    [fieldName]: "error text",
  },
  handout: {
    ...
  },
}
```

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/hooks.js
rayzhou-bit: yes

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/CollapsibleFormWidget.jsx
rayzhou-bit: ty!

src/editors/containers/VideoEditor/components/VideoEditorModal.jsx
connorhaugh: Ok yeah So that makes more sense -- we just want to be sure we aren't over-specifying the shape in places we don't have to.

connorhaugh (approved):
I'm ok with this approach, and after staring at the code for a while, I think I agree with it. For next time, when attempting to implement something like this, it might be a great time to use a React Context [](https://reactjs.org/docs/context.html) to avoid passing `error` at every level, so we have one central error object.

PR #108, opened 2022-08-30
feat: change image absolute urls to relative urls

KristinAoki:
This PR fixes the image src's being absolute URLs instead of relative URLs. This PR also adds functionality to convert all image src's to `/static/` links to match expected URL format in studio for course re-runs and export/import functionality. New tests added to test the URL change and old tests updated to match variable changes.

JIRA Ticket: [TNL-10104](https://2u-internal.atlassian.net/browse/TNL-10104)

### DIFF
--- a/src/editors/containers/EditorContainer/hooks.js
+++ b/src/editors/containers/EditorContainer/hooks.js
@@ -11,11 +11,40 @@ export const {
   saveBlock,
 } = appHooks;
 
+export const setAssetToStaticUrl = (images, getContent) => {
+  let content = getContent();
+  const imageUrls = [];
+  const imgsArray = Object.values(images);
+  imgsArray.forEach(image => {
+    imageUrls.push({ portableUrl: image.portableUrl, displayName: image.displayName });
+  });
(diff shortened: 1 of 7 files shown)


src/editors/containers/EditorContainer/hooks.js
```diff
@@ -11,11 +11,40 @@ export const {
   saveBlock,
 } = appHooks;
 
+export const setAssetToStaticUrl = (images, getContent) => {
```
connorhaugh, line 14: Can you add a comment here for future readers which summarizes the conversion? I think it would be something like:
`// For assets to remain usable across course instances, we convert their url to be course-agnostic. That for example, /course/assets/<asset hash>/ filename` gets converted to '/static/<etc>/<fileName>'

connorhaugh: Also might be worth to note that if we add an assets storage method for libraries, this would be something that would need to choose? I don't really know how far off assets are in libraries, but good to be future-proof.

connorhaugh (approved):
Looks good to me! Nice job

PR #112, opened 2022-09-08
feat: unescape alphanumerics in code Editor

connorhaugh:
For: https://2u-internal.atlassian.net/browse/TNL-10100
 This adds a regex on the loading of content in the code editor to un-escape all required characters common html characters sourced from [here](https://www.w3.org/wiki/Common_HTML_entities_used_for_typography).
The thing is people might get in a fuss if they want to see the unescaped characters. Consequently, I have added a toggle to the raw editor view which escapes all html literals.


src/editors/containers/TextEditor/components/CodeEditor/index.jsx
```diff
@@ -26,17 +35,51 @@ export const hooks = {
...
+      (escapedKeyword) => {
+        // eslint-disable-next-line no-cond-assign
+        for (let next; (next = text.indexOf(alphanumericMap[escapedKeyword], pos)) > -1;) {
+          changes.push({ from: next, to: next + 1, insert: `&${escapedKeyword};` });
+          pos = next + 1;
+        }
+      },
+    );
+    ref.current.dispatch({ changes });
+    btnRef.current.setAttribute('disabled', 'disabled');
```
KristinAoki, line 57: Why are you disabling the button? Would we want the button to toggle instead? Also when it is disabled the button is no longer readable. I can see that it is still there but it is very difficult to read.

connorhaugh: Because this is an activity you can only do once. You don't want to exscape string literals witch, as that results in for example `&ldquo` -> `&ampldquo` (as it reads the ampersand), which isn't what we want.

KristinAoki: Do you want to remove the button from view instead of disabling it?

connorhaugh: Toggling is too much work for what our partners want (they either want one or the other.)

connorhaugh: Sure why not I'll do that?

connorhaugh: @KristinAoki I did that

KristinAoki (approved):

PR #113, opened 2022-09-08
feat: allow .ico image file types

KristinAoki:
This PR allows .ico images to be uploaded in the text editor. File input snapshot was updated to include new allowed file type and no new tests were added.

JIRA Ticket: [TNL-9952](https://2u-internal.atlassian.net/browse/TNL-9952)


connorhaugh (approved):
:shipit:

PR #116, opened 2022-09-21
feat: connect transcript child ui to api

KristinAoki:
This PR connects the UI and mocked api call for transcript child items to Studio API endpoints.

JIRA ticket: [TNL-9980](https://2u-internal.atlassian.net/browse/TNL-9980)

### DIFF
--- a/src/editors/containers/VideoEditor/components/VideoSettingsModal/components/TranscriptWidget/hooks.js
+++ b/src/editors/containers/VideoEditor/components/VideoSettingsModal/components/TranscriptWidget/hooks.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import { actions, selectors } from '../../../../../../data/redux'
+import { thunkActions } from '../../../../../../data/redux';
 
 export const transcriptLanguages = (transcripts) => {
   const languages = [];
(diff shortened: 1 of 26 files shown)


src/editors/containers/VideoEditor/components/VideoSettingsModal/components/TranscriptWidget/index.jsx
```diff
@@ -35,21 +38,12 @@ export const TranscriptWidget = ({
...
-      show: true,
-    },
-  };
-  const upload = {
-    error: {
-      dismiss: () => { console.log('dismiss'); },
-      show: true,
-    },
-  };
+  const fileInput = hooks.fileInput({ onAddFile: '' });
```
connorhaugh, line 45: We'll need to add an onAddFile action here in order to integrate our work

connorhaugh:
For those curious: the reason Kristin is only using the upload/delete endpoints is because the replace endpoint actually: 

    """
    Downloads subtitles from youtube and replaces edx transcripts in DS with youtube ones.
    """

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/TranscriptWidget/hooks.js
```diff
@@ -11,19 +13,57 @@ export const transcriptLanguages = (transcripts) => {
...
   const addFile = (e) => {
-    const selectedFile = e.target.files[0];
-    console.log(selectedFile);
+    if (typeof onAddFile === 'function') {
+      onAddFile(e);
+    } else {
+      const selectedFile = e.target.files[0];
+      dispatch(thunkActions.video.uploadTranscript({
+        filename: selectedFile.name,
+        language: 'esp',
```
connorhaugh, line 46: we might want to add something more sophisticated here as to prevent collision-- as to prevent lang overlap we can grab from selector openLanguages and grab the first one.

connorhaugh: I think that would be part of the laundry list of little things we can do to later improve the feature.

src/editors/data/redux/thunkActions/video.js
```diff
@@ -10,7 +14,66 @@ export const saveVideoData = () => () => {
   // dispatch(requests.saveBlock({ });
 };
 
-export default {
+export const uploadTranscript = ({ language, filename, file }) => (dispatch, getState) => {
+  // TODO: Prevent the addition of dulicate keys by setting language here
```
connorhaugh, line 18: it might be easier to do that lang selection if empty here.

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/TranscriptWidget/index.jsx
connorhaugh: Wait, now that I've read the file, can we follow the same pattern of passing in an action for both, as not to confuse the reader of our code?

connorhaugh:
Closing in favor of the combination PR, which has been merged.

PR #117, opened 2022-09-22
Feat:  full transcript widget

connorhaugh:
This ticket resolves https://2u-internal.atlassian.net/browse/TNL-9822 

Left to accomplish or split out:

- [ ] Files that are too large should error out with pertinent message. (For Later)
- [x] Files should delete only on click of confirm delete
- [x] Use full lang names on collapse (For Later)
- [x] error up propagation/ widget collapsing on error.
- [x] convert download button from variable to thunkAction


src/editors/containers/VideoEditor/components/VideoSettingsModal/components/TranscriptWidget/index.jsx
```diff
@@ -58,23 +53,27 @@ export const TranscriptWidget = ({
       title="Transcript"
```
KristinAoki, line 53: need to change to `{intl.formatMessage(messages.titleLabel)}`

src/editors/data/redux/video/selectors.js
```diff
@@ -1,16 +1,20 @@
...
+import * as AppSelectors from '../app/selectors';
+import { downloadVideoTranscriptURL } from '../../services/cms/urls';
 
 const stateKeys = keyStore(initialState);
 
 export const video = (state) => state.video;
 
 export const simpleSelectors = [
   stateKeys.videoSource,
+  stateKeys.videoId,
```
rayzhou-bit, line 17: Is this specifically edx video id?

src/editors/data/services/cms/api.js
```diff
@@ -29,6 +29,35 @@ export const apiMethods = {
...
+  }) => {
+    const data = new FormData();
+    data.append('file', transcript);
+    data.append('edx_video_id', videoId);
+    data.append('language_code', language);
+    data.append('new_language_code', language);
+    return post(
+      urls.videoTranscripts({ studioEndpointUrl, blockId }),
+      data,
+    );
```
rayzhou-bit, line 59: Can you upload a transcript without an edx_video_id?

src/editors/data/redux/video/selectors.js
KristinAoki: yes, it is needed for the transcript API calls

src/editors/data/services/cms/api.js
KristinAoki: I don't think so.

rayzhou-bit (commented):
What happens when there is no edx_video_id?

KristinAoki:
> What happens when there is no edx_video_id? Should we display widget still?

After more testing with an empty string, undefined variable, and random string, `edx_video_id` is required for the transcripts to be mapped to the correct video. Transcripts are mapped to whatever value is provided for `edx_video_id`.

rayzhou-bit:
It might be time to pull Jon in for design questions.

Should the user be allowed to attempt a transcript upload without an edx video id? Based on everything else so far, I imagine Jon would say yes and that we should display an error.

What happens to transcript widget if user changes the edx video id? From playing around on stage, I noticed that the transcript seems to be tied directly to the video file and not the video block.

connorhaugh:
@rayzhou-bit so when a transcript is uploaded, it creates a video_id if there is none presently associated with the xblock. changing the video id to another will still keep the transcripts. 
This is the first request (no id)
Screen Shot 2022-09-27 at 10 39 57 AM
this is the response (id created).
Screen Shot 2022-09-27 at 10 42 18 AM
We can use this to set the video id if there is none set locally in the MFE. maybe?

rayzhou-bit:
Just to confirm, if we call this and videoId is empty, backend will create an edx video id, connect it to the uploaded video of the block and the transcript will be connected to that id?
image

src/editors/data/redux/thunkActions/video.js
```diff
@@ -24,12 +24,20 @@ export const uploadTranscript = ({ language, filename, file }) => (dispatch, get
...
-      },
-    })),
+    onSuccess: (response) => {
+      dispatch(actions.video.updateField({
+        transcripts: {
+          ...transcripts,
+          [lang]: { filename },
+        },
+      }));
+      if (selectors.video.videoId(state) === '') {
```
connorhaugh, line 34: @rayzhou-bit this is where that code would occur, you are looking at the requests code I believe.

rayzhou-bit: Oh derp. I missed this

rayzhou-bit (approved):

PR #118, opened 2022-09-22
feat: videosource to backend

rayzhou-bit:
https://2u-internal.atlassian.net/browse/TNL-10011

The code change allows the frontend to send widget information to the backend in the correct format. While this focuses on video source widget, it expands on the current framework to send information from other widgets as well.


src/editors/data/services/cms/api.js
```diff
@@ -39,13 +39,43 @@ export const apiMethods = {
     if (blockType === 'html') {
       return {
         category: blockType,
-        couseKey: learningContextId,
+        courseKey: learningContextId,
```
rayzhou-bit, line 42: Pretty sure this was a typo

src/editors/containers/VideoEditor/index.jsx
```diff
@@ -24,11 +29,22 @@ export default function VideoEditor({
...
+};
 
 VideoEditor.defaultProps = {
   onClose: null,
+  videoSettings: null,
 };
 VideoEditor.propTypes = {
   onClose: PropTypes.func,
+  // redux
+  videoSettings: PropTypes.shape({}),
```
connorhaugh, line 41: you can have two different shapes here (loaded/unloaded)

src/editors/data/redux/video/selectors.js
```diff
@@ -23,6 +23,38 @@ export const simpleSelectors = [
   stateKeys.licenseDetails,
 ].reduce((obj, key) => ({ ...obj, [key]: state => state.video[key] }), {});
 
+export const videoSettings = createSelector(
```
connorhaugh, line 26: We'll need to, after we merge transcripts, refactor this

src/editors/data/services/cms/api.js
```diff
@@ -39,13 +39,43 @@ export const apiMethods = {
...
+        display_name: title,
+        id: blockId,
+        metadata: {
+          display_name: title,
+          download_video: content.allowVideoDownloads,
+          edx_video_id: edxVideoId,
+          html5_sources: html5Sources,
+          youtube_id_1_0: youtubeId,
+          download_track: content.allowTranscriptDownloads,
+          track: '', // TODO Downloadable Transcript URL. Backend expects a file name, for example: "something.srt"
```
connorhaugh, line 70: How does this mesh with the work we did @KristinAoki

src/editors/data/redux/video/selectors.js
rayzhou-bit: The idea of that selector would be to grab all the values we would want to send. I would want to at least re-evaluate it after we finish all the widgets for video editor since it seems like we are changing the data shape as we learn new things about video.

src/editors/containers/VideoEditor/index.jsx
rayzhou-bit: This covers both loaded and unloaded

connorhaugh: Sorry, what I meant was, this does pass, but gives no information to the reader what kind of prop this is, instead try:

`videoSettings: PropTypes.oneOfType([Proptypes.shape({}), Proptypes.Shape({key1: Proptypes.something, key2... })])`

rayzhou-bit: Oh I see. I updated it to reflect what is in mockVideoData

connorhaugh (approved):
👍 Ty

PR #124, opened 2022-10-05
fix: refactor getContent to take Param dispatch

connorhaugh:


src/editors/containers/EditorContainer/hooks.js
```diff
@@ -11,46 +11,13 @@ export const {
...
-
 export const handleSaveClicked = ({ dispatch, getContent, validateEntry }) => {
   const destination = useSelector(selectors.app.returnUrl);
   const analytics = useSelector(selectors.app.analytics);
-  const images = useSelector(selectors.app.images);
+
   return () => saveBlock({
     analytics,
-    content: setAssetToStaticUrl(images, getContent),
+    content: getContent({ dispatch }),
```
KristinAoki, line 20: will this dispatch get confused with the one passed into the function?

connorhaugh: this dispatch is the one passed into this function. The original dispatch for this function is created in the EditorContainer

PR #125, opened 2022-10-06
Feat  video source integration

rayzhou-bit:


src/editors/containers/VideoEditor/components/VideoSettingsModal/components/ErrorSummary.jsx
```diff
@@ -1,40 +1,27 @@
...
-  },
-};
-ErrorSummary.propTypes = {
-  error: PropTypes.node,
+export const ErrorSummary = () => {
+  const error = React.useContext(ErrorContext);
+  return (
+    <Alert
+      icon={Info}
+      show={!Object.values(error).every(val => Object.keys(val[0]).length === 0)}
```
KristinAoki, line 14: I think it might be better to separate the function inside of `every()` as its own and call it here. It is a little difficult to read what is happening here.  It would also allow for a bit more variable naming to make the readability clearer.

KristinAoki: I would also change the length check from `=== 0` to `!==0` and you won't have to put `!` at the beginning of the line.

KristinAoki (changes requested):

src/editors/data/redux/video/selectors.js
```diff
@@ -47,7 +50,19 @@ export const getTranscriptDownloadUrl = createSelector(
 );
 
 export const videoSettings = createSelector(
-  Object.values(module.simpleSelectors),
+  [
```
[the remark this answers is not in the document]
rayzhou-bit: Originally I wanted to create a selector that took every simpleSelector. When videoId was added, we were getting an error because the values didn't line up anymore. I changed this selector to grab just these values since this should be all we need for saving.

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/ErrorSummary.jsx
KristinAoki: looks good!

src/editors/containers/VideoEditor/components/VideoSettingsModal/__snapshots__/ErrorSummary.test.jsx.snap
```diff
@@ -0,0 +1,45 @@
...
+    <FormattedMessage
+      defaultMessage="Please check your entries and try again."
+      description="Body of validation error."
+      id="authoring.videoeditor.validate.error.body"
+    />
+  </p>
+</Alert>
+`;
+
+exports[`ErrorSummary render snapshots: renders as expected when there are no errors 1`] = `
```
KristinAoki, line 25: Shouldn't this snapshot return nothing or at least the Alert param `show` should equal `false`

src/editors/containers/VideoEditor/components/VideoSettingsModal/ErrorSummary.jsx
```diff
@@ -0,0 +1,31 @@
...
+    <Alert
+      icon={Info}
+      show={Object.values(error).every(module.hasError)}
+      variant="danger"
+    >
+      <Alert.Heading>
+        <FormattedMessage {...messages.validateErrorTitle} />
+      </Alert.Heading>
+      <p>
+        <FormattedMessage {...messages.validateErrorBody} />
```
KristinAoki, line 25: Will this always be a fixed message or will it change based on the errors from the components? Like will it tell you which widgets have the errors? If not, does it make sense for the values to be booleans instead of objects?

KristinAoki (changes requested):
It appears that error summary tests are always returning true. i also noticed that in the transcript test snapshots that `isError` was always set to true, but should only be true when uploadError and deleteError are true.

src/editors/containers/VideoEditor/components/VideoSettingsModal/ErrorSummary.jsx
rayzhou-bit: Based on the figma, it would always be a fixed message. I figured it should just pull in the error context to evaluate the boolean here. Normally, we add a hooks file, but there's not that much logic here.

src/editors/containers/VideoEditor/components/VideoSettingsModal/__snapshots__/ErrorSummary.test.jsx.snap
rayzhou-bit: I fixed this. Realized that the logic I had would have shown alert only if every widget had an error.

KristinAoki (approved):
LGTM

PR #126, opened 2022-10-11
feat: limit xblock title display 

KristinAoki:
Previously titles could have an indefinite length that would run off the screen and block the modal close button. This PR adds a truncated view to xblock titles when not editing. In the editing view the input is saved on the check button click and discarded on the close button click.  

JIRA Ticket: [TNL-9916](https://2u-internal.atlassian.net/browse/TNL-9916)


src/editors/containers/EditorContainer/components/TitleHeader/EditableHeader.jsx
```diff
@@ -1,25 +1,26 @@
...
   handleChange,
   updateTitle,
   handleKeyDown,
   inputRef,
   localTitle,
+  cancelEdit,
 }) => (
   <Form.Group>
     <Form.Control
+      style={{ 'padding-inline-end': 'calc(1rem + 84px)' }}
```
connorhaugh, line 17: nit: for next time: https://mdbootstrap.com/docs/react/utilities/spacing/ we use classNames to reference bootstrap styles, in this case. I'm going to say this is ok for now, because this is something which improves performance so minisculely.

connorhaugh (approved):
Looks good, just one minor nit

src/editors/containers/EditorContainer/components/TitleHeader/EditableHeader.jsx
KristinAoki: The padding only allows for up to the value 5 which was not enough for two buttons inside the input. So I had to include this work around.

PR #130, opened 2022-10-19
feat: update setAssetToStaticUrl to include all assets

KristinAoki:
This PR expands the “static image URL” functionality that was built for the new text editor to be applied to ALL files and uploads assets. This ensures that all files linked in the visual and raw text editor will update with static URLs when saved and allow them to correctly update on course re-runs or export/imports.

JIRA Ticket: [TNL-10150](https://2u-internal.atlassian.net/browse/TNL-10150)


connorhaugh (approved):
Thanks, this look great, and will be much cleaner

PR #131, opened 2022-10-20
feat: duration widget for video settings. TNL-9823.

kenclary:


src/editors/containers/VideoEditor/components/VideoSettingsModal/components/DurationWidget.jsx
```diff
@@ -22,8 +23,23 @@ export const DurationWidget = () => {
...
+      return `Starts at ${durationFromValue(startTime)}`; // TODO internationalize
+    }
+    const total = stopTime - (startTime || 0);
+    return `Total: ${durationFromValue(total)}`; // TODO internationalize
+  };
+
   return (
-    <CollapsibleFormWidget title="Duration">
+    <CollapsibleFormWidget
+      title="Duration"
```
KristinAoki, line 40: internationalization

src/editors/data/services/cms/mockApi.js
```diff
@@ -123,7 +123,7 @@ export const fetchAssets = ({ learningContextId, studioEndpointUrl }) => mockPro
 
 // eslint-disable-next-line
 export const allowThumbnailUpload = ({ studioEndpointUrl }) => mockPromise({
-  data: true,
+  data: true || studioEndpointUrl, // shut up, lint,
```
KristinAoki, line 126: This isn't needed, mine is doesn't throw the lint error

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/DurationWidget.jsx
```diff
@@ -22,8 +23,23 @@ export const DurationWidget = () => {
...
+      subtitle={getTotalLabel(duration.formValue.startTime, duration.formValue.stopTime)}
+    >
       <FormGroup size="sm">
         <div>
           <FormControl
```
KristinAoki, line 45: refer to the mock-ups in ticket for the helper text that should be under each of the fields. Also i think that these need to be number inputs versus text inputs.

KristinAoki (changes requested):
Addition to the comments that I left, there are a few ACs from the ticket that aren't reflected currently.

- When the fields are cleared they are changed to NA:NA:NA instead of 00:00:00
- The end time shouldn't be able to before the start time and vice versa
- Video must be at least a second long (imagine that this check would only happen is a non-default value is input)

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/DurationWidget.jsx
kenclary: the inputs are text, so they can display '00:00:00' format. duration.js handles the conversion between that and actual numeric values (milliseconds)

KristinAoki (changes requested):

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/DurationWidget.jsx
```diff
@@ -5,48 +5,85 @@ import { useDispatch } from 'react-redux';
...
+
   return (
-    <CollapsibleFormWidget title="Duration">
+    <CollapsibleFormWidget
+      title={(<FormattedMessage {...messages.durationTitle} />)}
+      subtitle={getTotalLabel(duration.formValue.startTime, duration.formValue.stopTime)}
+    >
       <FormGroup size="sm">
-        <div>
+        <div className="d-inline-block">
```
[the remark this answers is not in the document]
kenclary: floatingLabel does not appear to match the behavior of the mockups (in which it is always "floated" to the input border). Also, even regular floatingLabel behavior may be broken in respects to the code in duration.js.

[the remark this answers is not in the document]
KristinAoki: The floatingLabel is always at the top border if there is something in the input box. If you are looking at the behavior in the gallery view, it is a bit weird. If you go to devstack you see that it behaves as expected.

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/DurationWidget.jsx
```diff
@@ -3,50 +3,85 @@ import { useDispatch } from 'react-redux';
...
           />
-        </div>
+          <FormControl.Feedback>
+            <FormattedMessage {...messages.durationHint} />
+          </FormControl.Feedback>
+        </FormGroup>
         <div className="mt-4">
-          Total: {duration.formValue.total}
+          {getTotalLabel(duration.formValue.startTime, duration.formValue.stopTime)}
         </div>
```
KristinAoki, line 76: Move this outside of the row

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/DurationWidget.jsx
```diff
@@ -3,50 +3,85 @@ import { useDispatch } from 'react-redux';
...
   return (
-    <CollapsibleFormWidget title="Duration">
-      <FormGroup size="sm">
-        <div>
+    <CollapsibleFormWidget
+      title={intl.formatMessage(messages.durationTitle)}
+      subtitle={getTotalLabel(duration.formValue.startTime, duration.formValue.stopTime)}
+    >
+      <FormattedMessage {...messages.durationDescription} />
+      <Row>
```
KristinAoki, line 51: Add top padding to prevent overlap between duration description and the start time floating label

src/editors/data/redux/thunkActions/video.js
```diff
@@ -32,9 +33,9 @@ export const loadVideoData = () => (dispatch, getState) => {
     allowTranscriptDownloads: rawVideoData.download_track,
     showTranscriptByDefault: rawVideoData.show_captions,
     duration: { // TODO duration is not always sent so they should be calculated.
-      startTime: rawVideoData.start_time,
-      stopTime: rawVideoData.end_time,
-      total: null, // TODO can we get total duration? if not, probably dropping from widget
+      startTime: valueFromDuration(rawVideoData.start_time || '00:00:00'),
+      stopTime: valueFromDuration(rawVideoData.end_time || '00:00:00'),
+      total: 0, // TODO can we get total duration? if not, probably dropping from widget
```
KristinAoki, line 38: remove this todo

src/editors/data/redux/thunkActions/video.js
```diff
@@ -32,9 +33,9 @@ export const loadVideoData = () => (dispatch, getState) => {
     allowTranscriptDownloads: rawVideoData.download_track,
     showTranscriptByDefault: rawVideoData.show_captions,
     duration: { // TODO duration is not always sent so they should be calculated.
```
KristinAoki, line 35: remove todo

KristinAoki (changes requested):
Some small ui changes then it should be good

src/editors/data/redux/thunkActions/video.js
kenclary: this might be a future enhancement (getting the real video total from the backend, which we don't have yet), so I'm tempted to leave it in.

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/DurationWidget.jsx
kenclary: is there a paragon standard class for adding space like this?

KristinAoki: i think you can add to row `className="mt3"` and that should do it

kenclary: mt-4, in fact (applied to the Row below)

KristinAoki (approved):

PR #137, opened 2022-11-03
feat: remove handout widget for content libraries

KristinAoki:
Content libraries do not have asset pages, so the ability to upload handouts is not possible. As a result, the handout widget is not displayed if it is a library block.


rayzhou-bit (approved):
lgtm!

PR #138, opened 2022-11-03
feat: remove thumbnail widget from content libraries

KristinAoki:
Add conditional that checks if the block is from a content library. Content libraries do not host their own videos. Therefore any edx-hosted video present in a content library originates from another course. When trying to upload a new thumbnail it errors out because it is trying to find the video in the library. To prevent this error, the thumbnail widget is not visible to users in content libraries.


rayzhou-bit (approved):
lgtm

PR #141, opened 2022-11-09
Feat  improve transcript flow

connorhaugh:
https://2u-internal.atlassian.net/browse/TNL-10199 
Rewrite of Transcripts widget to improve flow.

TODO: rewrite tests


src/editors/containers/EditorContainer/components/TitleHeader/index.jsx
```diff
@@ -45,13 +45,10 @@ export const TitleHeader = ({
   }
   return (
     <div className="d-flex">
-      <Truncate>
```
connorhaugh, line 48: I realize I need to take this out

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/TranscriptWidget/index.jsx
```diff
@@ -42,13 +82,12 @@ export const TranscriptWidget = ({
...
-  const fileInput = hooks.fileInput({ onAddFile: hooks.addFileCallback({ dispatch: useDispatch() }) });
-  const hasTranscripts = hooks.hasTranscripts(transcripts);
+  const fullTextLanguages = module.hooks.transcriptLanguages(transcripts);
+  const hasTranscripts = module.hooks.hasTranscripts(transcripts);
   return (
     <CollapsibleFormWidget
       isError={Object.keys(error).length !== 0}
-      subtitle={languagesArr}
+      subtitle={fullTextLanguages}
       title="Transcript"
```
KristinAoki, line 91: internationalize

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/TranscriptWidget/index.jsx
```diff
@@ -67,10 +106,10 @@ export const TranscriptWidget = ({
         {hasTranscripts ? (
 
           <Form.Group className="mt-4.5">
-            { Object.entries(transcripts).map(([language, value]) => (
-              <TranscriptListItem
+            { transcripts.map((language, index) => (
+              <Transcript
```
KristinAoki, line 110: Add some padding around this component. There is no spacing when the delete card appears and the transcript below it.

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/TranscriptWidget/TranscriptActionMenu.jsx
```diff
@@ -0,0 +1,80 @@
...
+      <FileInput fileInput={input} acceptedFiles=".srt" />
+    </Dropdown>
+  );
+};
+
+TranscriptActionMenu.propTypes = {
+  index: PropTypes.number.isRequired,
+  language: PropTypes.string.isRequired,
+  // redux
+  launchDeleteConfirmation: PropTypes.func.isRequired,
```
KristinAoki, line 68: is this function actually in redux? if so update in props.

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/TranscriptWidget/Transcript.jsx
```diff
@@ -0,0 +1,104 @@
...
+export const Transcript = ({
+  index,
+  language,
+  // redux
+  deleteTranscript,
+}) => {
+  const { inDeleteConfirmation, launchDeleteConfirmation, cancelDelete } = module.hooks.setUpDeleteConfirmation();
+  return (
+    <>
+      {inDeleteConfirmation
```
KristinAoki, line 46: After deleting a transcript the card does not go away (e.i. the state is not updating to false). How is the state attached to a specific card?

src/editors/data/redux/thunkActions/video.js
```diff
@@ -197,31 +212,29 @@ export const uploadHandout = ({ file }) => (dispatch) => {
...
 
-export const uploadTranscript = ({ language, filename, file }) => (dispatch, getState) => {
+export const uploadTranscript = ({ language, file }) => (dispatch, getState) => {
   const state = getState();
   const { transcripts, videoId } = state.video;
-  let lang = language;
-  if (!language) {
-    [[lang]] = selectors.video.openLanguages(state);
-  }
+  console.log({ roastBeast: transcripts });
```
KristinAoki: remove

KristinAoki (changes requested):

src/editors/data/services/cms/api.js
```diff
@@ -97,6 +111,7 @@ export const apiMethods = {
       };
     }
     if (blockType === 'video') {
+      // found da bug! It is here
```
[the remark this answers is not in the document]
connorhaugh: hahahaha no way edx_video_id was not being saved to the block if it was set by uploading a transcript

src/editors/data/services/cms/api.js
```diff
@@ -68,12 +81,13 @@ export const apiMethods = {
     transcript,
     videoId,
     language,
+    newLanguage = null,
```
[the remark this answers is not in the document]
connorhaugh: it is for when you are changing the language associated with a transcript.

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/TranscriptWidget/Transcript.jsx
connorhaugh: > After deleting a transcript the card does not go away.
Ahh! It in fact does, if there is no card below it, if there is, that one takes the index and keeps the state. I will fix this.

src/editors/data/redux/thunkActions/video.js
```diff
@@ -232,31 +245,44 @@ export const deleteTranscript = ({ language }) => (dispatch, getState) => {
...
+      console.log({ updatedTranscripts });
+      dispatch(actions.video.updateField({ transcripts: updatedTranscripts }));
+    },
+  }));
+};
+
+export const updateTranscriptLanguage = ({ newLanguageCode, languageBeforeChange }) => (dispatch, getState) => {
+  const state = getState();
+  const { video: { transcripts, videoId } } = state;
+  selectors.video.getTranscriptDownloadUrl(state);
```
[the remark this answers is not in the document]
connorhaugh: > Is there a reason you are pulling out the video object out from the state instead of `const {transcripts, videoId} = state.video?

No there is no difference between the two.

why are you using selectors when you have the full state provided in the previous line?
Because the getTranscriptDownloadUrl is a selector which is dependent on state values, but itself not a state value. you're not wrong that I could make them all match, but that would require more mocking in the tests and I'm lazy

src/editors/data/redux/thunkActions/video.js
```diff
@@ -232,31 +245,44 @@ export const deleteTranscript = ({ language }) => (dispatch, getState) => {
...
 
 export const replaceTranscript = ({ newFile, newFilename, language }) => (dispatch, getState) => {
   const state = getState();
-  const { transcripts, videoId } = state.video;
+  const { videoId } = state.video;
   dispatch(requests.deleteTranscript({
     language,
     videoId,
     onSuccess: () => {
-      const updateTranscripts = {};
```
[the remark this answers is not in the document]
connorhaugh: I've solved this by not double adding transcripts on upload

src/editors/sharedComponents/FileInput/index.jsx
```diff
@@ -1,12 +1,28 @@
...
+    }
+  };
+  return {
+    click,
+    addFile,
+    ref,
+  };
+};
+
+export const FileInput = ({ fileInput: hook, acceptedFiles }) => (
```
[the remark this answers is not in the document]
connorhaugh: Note that they don't use the `fileInput` hook so it doesn't effect them.

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/TranscriptWidget/index.jsx
connorhaugh: Done! I added it in the delete card

rayzhou-bit:
If you upload a transcript, change the language and try to download, you will get an error. However, if you save after changing the language and then reload, you can download the transcript as the new saved language. Getting the error isn't great but I don't think this is a blocker.

Doing the same as above but deleting instead of downloading will not delete the transcript. Not a blocker either.

You can select an already uploaded transcript language.

I do like that you can change the language after upload. It's a nice way for users to quickly correct things. You could solve the issues above by making the dropdown inactive though. That would basically be legacy behavior.

src/editors/data/redux/thunkActions/video.js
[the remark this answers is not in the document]
KristinAoki: I just tested this and when I replace a transcript the original one stays and a new one appears. However, when i go to delete the first one all of them are deleted.

KristinAoki:
> If you upload a transcript, change the language and try to download, you will get an error. However, if you save after changing the language and then reload, you can download the transcript as the new saved language. Getting the error isn't great but I don't think this is a blocker.
> 
> Doing the same as above but deleting instead of downloading will not delete the transcript. Not a blocker either.
I did not get these errors when I tried to download or delete transcript after changing the language. For the download, it does download, but with the original language code.
> You can select an already uploaded transcript language.
Confirmed this

KristinAoki (commented):
Once the code coverage increases for:
- src/editors/data/redux/thunkActions/requests.js
- src/editors/data/redux/thunkActions/video.js
- src/editors/data/services/cms/api.js
the PR will be ready to be approved

connorhaugh:
https://app.codecov.io/github/edx/frontend-lib-content-components/commit/a097d55f99658742d09b2fa0f060a05dfa88fb50 
Here you go!

KristinAoki (approved):

PR #147, opened 2022-11-28
fix: length error on undefined fallback videos

KristinAoki:
JIRA Ticket: [TNL-10227](https://2u-internal.atlassian.net/browse/TNL-10227)

This PR addresses the error message "an unexpected error occurred. Please click the button below to refresh the page.". When course is created with xml, it does not define `html5Sources` as an empty list. This PR updates the conditional that sets `fallbackVideos` to `['', '']` so that it also check for undefined values.


connorhaugh (approved):
Looks good to me!  note that the statement `if(fallbackvideos?.length)` also works as `undefined` and `0` are falsy but in this situation optional chaining is more confusing than it is worth.

rayzhou-bit (approved):

PR #150, opened 2022-11-30
feat: make delete and info icons outlines

connorhaugh:
Solves: https://2u-internal.atlassian.net/browse/TNL-10239

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -34,7 +34,7 @@
       "devDependencies": {
         "@edx/frontend-build": "^11.0.2",
         "@edx/frontend-platform": "2.4.0",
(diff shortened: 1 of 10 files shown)


KristinAoki (commented):
All the outlines look good! One thing to note is that when you have over the `DeleteOutline` icon, it does not change to a white outline so it is a black outline on a navy background. Should this be filed as a bug ticket for paragon?

connorhaugh:
Good Note I'll put this to jfay to reveiw

connorhaugh:
Blocked on https://github.com/openedx/paragon/issues/1814

PR #161, opened 2022-12-07
feat: header overlap content when scrolling down

rayzhou-bit:
https://2u-internal.atlassian.net/browse/TNL-10234

This change will keep header above the content when user scrolls down. TinyMCE sticky toolbar in the text editor is kept at the appropriate height.

### DIFF
--- a/src/editors/containers/EditorContainer/__snapshots__/index.test.jsx.snap
+++ b/src/editors/containers/EditorContainer/__snapshots__/index.test.jsx.snap
@@ -2,7 +2,9 @@
 
 exports[`EditorContainer component render snapshot: initialized. enable save and pass to header 1`] = `
 <div>
-  <ModalDialog.Header>
+  <ModalDialog.Header
+    className="border-bottom zindex-10"
+  >
     <ModalDialog.Title>
       <div
         style={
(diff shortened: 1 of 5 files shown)


src/editors/containers/EditorContainer/__snapshots__/index.test.jsx.snap
```diff
@@ -2,7 +2,9 @@
 
 exports[`EditorContainer component render snapshot: initialized. enable save and pass to header 1`] = `
 <div>
-  <ModalDialog.Header>
+  <ModalDialog.Header
+    className="border-bottom zindex-10"
```
KristinAoki, line 6: Can you make it a drop shadow instead of border line to match the expected design in paragon?

src/editors/containers/EditorContainer/index.jsx
```diff
@@ -20,7 +20,7 @@ export const EditorContainer = ({
   const handleCancelClicked = hooks.handleCancelClicked({ onClose });
   return (
     <div>
-      <ModalDialog.Header>
+      <ModalDialog.Header className="border-bottom zindex-10">
```
KristinAoki, line 23: Refer to previous comment

KristinAoki (changes requested):

KristinAoki (approved):

PR #162, opened 2022-12-09
feat: parseTranscript to work with all languages

rayzhou-bit:
https://2u-internal.atlassian.net/browse/TNL-10281

Video Settings run into an error when opening while preferred language is not english. This is because we are expecting english when parsing transcript files. This change aims to parse transcript files regardless of the user's preferred language.


connorhaugh (approved):
+1 to making my hacks less hacky tysm

PR #164, opened 2022-12-09
feat: separate video id and url fields into two

KristinAoki:
JIRA Ticket: [TNL-10274](https://2u-internal.atlassian.net/browse/TNL-10274)

This PR separates the Video ID or URL field into separate fields. If a user happens to have both saved to the block it will separate them into the respective fields. Additionally when editing, if a user tries to but a video ID in the URL field is will adjust it to be in the correct field and vice versa.

### DIFF
--- a/src/editors/containers/VideoEditor/components/VideoSettingsModal/components/ThumbnailWidget/index.test.jsx
+++ b/src/editors/containers/VideoEditor/components/VideoSettingsModal/components/ThumbnailWidget/index.test.jsx
@@ -3,7 +3,9 @@ import { shallow } from 'enzyme';
 
 import { formatMessage } from '../../../../../../../testUtils';
 import { selectors } from '../../../../../../data/redux';
+import { isEdxVideo } from '../../../../../../data/services/cms/api';
 import { ThumbnailWidget, mapStateToProps, mapDispatchToProps } from '.';
+import { before } from 'lodash-es';
 
 jest.mock('react', () => ({
   ...jest.requireActual('react'),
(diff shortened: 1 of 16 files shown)


src/editors/containers/VideoEditor/components/VideoSettingsModal/components/ThumbnailWidget/index.test.jsx
```diff
@@ -52,22 +52,22 @@ describe('ThumbnailWidget', () => {
     });
     test('snapshots: renders as expected with a thumbnail provided', () => {
       expect(
-        shallow(<ThumbnailWidget {...props} thumbnail="sOMeUrl" videoType="edxVideo" />),
+        shallow(<ThumbnailWidget {...props} thumbnail="sOMeUrl" videoId="e0466f70-9fb0-40d0-bcd0-3a3f6e9fa890" />),
```
connorhaugh, line 55: nit: 
```suggestion
        shallow(<ThumbnailWidget {...props} thumbnail="sOMeUrl" videoId="sOmEID" />),
```
and then we mock out isEdxVideo

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/VideoSourceWidget/__snapshots__/index.test.jsx.snap
```diff
@@ -10,14 +10,39 @@ exports[`VideoSourceWidget snapshots snapshots: renders as expected with default
...
-        floatingLabel="Video ID or URL"
+        floatingLabel="Video ID"
         onBlur={[Function]}
         onChange={[MockFunction]}
         value=""
       />
+      <Component
+        className="text-primary-300 my-3"
+      >
+        <FormattedMessage
```
connorhaugh, line 21: Screen Shot 2022-12-12 at 9 41 03 AM
The bottom padding looks odd to me here, (in theory the label should be closer to the textinput it refers to?) But if it matches mocks all cool with that, I can't seem to find them.

connorhaugh:
it works well on my machine, just a couple of minor things.

connorhaugh (approved):
lgtm tysm!

PR #165, opened 2022-12-12
fix: add check if field is cleared

KristinAoki:
This is a follow on[ PR to separating video ID and video URL into two fields](https://github.com/openedx/frontend-lib-content-components/pull/164). In the previous implementation and user could clear one of the fields and it would end up clearing both. Now only the expected field is cleared onBlur.


connorhaugh (approved):
LGTM

PR #168, opened 2022-12-14
feat: add problem type preview

KristinAoki:
This PR adds the problem type preview on the problem select page.

JIRA Ticket: [TNL-10192](https://2u-internal.atlassian.net/browse/TNL-10192)


src/editors/containers/ProblemEditor/components/SelectTypeModal/content/Preview.jsx
```diff
@@ -1,37 +1,63 @@
...
+  intlShape,
+} from '@edx/frontend-platform/i18n';
+import messages from './messages';
+import { selectors } from '../../../../../data/redux';
 import { ProblemTypes } from '../../../../../data/constants/problem';
 
-const Preview = ({
+export const Preview = ({
+  // redux
   problemType,
```
connorhaugh, line 16: ah, so this will be wrong, as this preview should show the local `selected` value, not the problemType. This will make it so that the user sees the preview for the item they have selected. Having the redux value here will not work

src/editors/containers/ProblemEditor/components/SelectTypeModal/content/Preview.jsx
```diff
@@ -1,37 +1,63 @@
...
 };
 
-export default Preview;
+export const mapStateToProps = (state) => ({
+  problemType: selectors.problem.problemType(state),
+});
+
+export const mapDispatchToProps = {};
+
+export default injectIntl(connect(mapStateToProps, mapDispatchToProps)(Preview));
```
connorhaugh, line 63: because of the above, this won't have to be a connected component

connorhaugh (changes requested):
Just some issues with how you've set this up  let me know if you have questions

connorhaugh (approved):
Looks good to me -- note: @rayzhou-bit you will want to make sure you set a default problem type to be selected on page load

PR #170, opened 2022-12-19
feat: add advance problem select

KristinAoki:
JIRA Ticket: [TNL-10216](https://2u-internal.atlassian.net/browse/TNL-10216)

This PR adds the functionality for selecting advance problem types.


src/editors/containers/ProblemEditor/components/SelectTypeModal/SelectTypeWrapper/index.jsx
```diff
@@ -43,7 +43,7 @@ SelectTypeWrapper.defaultProps = {
   onClose: null,
 };
 SelectTypeWrapper.propTypes = {
-  selected: PropTypes.func.isRequired,
+  selected: PropTypes.string.isRequired,
```
rayzhou-bit, line 46: good catch

src/editors/containers/ProblemEditor/components/SelectTypeModal/content/ProblemTypeSelect.jsx
```diff
@@ -1,35 +1,43 @@
...
+import { ProblemTypes, ProblemTypeKeys, AdvanceProblemKeys } from '../../../../../data/constants/problem';
+import { FormattedMessage, injectIntl } from '@edx/frontend-platform/i18n';
+import messages from './messages';
 
 export const ProblemTypeSelect = ({
   selected,
   setSelected,
 }) => {
   const handleChange = e => setSelected(e.target.value);
+  const handleClick = () => setSelected('blankadvanced');
```
rayzhou-bit, line 13: use `AdvanceProblemKeys.BLANK`

src/editors/containers/ProblemEditor/components/SelectTypeModal/hooks.js
```diff
@@ -15,7 +16,11 @@ export const selectHooks = () => {
   };
 };
 
-export const onSelect = ( setProblemType, selected ) => () => {
+export const onSelect = ( setProblemType, selected, updateField ) => () => {
+  if (Object.values(AdvanceProblemKeys).includes(selected)) {
+    console.log(AdvanceProblems[selected]);
+    updateField({rawOLX: AdvanceProblems[selected].template})
```
rayzhou-bit, line 22: We should just create a new reducer called `setOlxTemplate` or something. Much more readable and direct.

rayzhou-bit (changes requested):
This looks great! Just have a couple thoughts and changes for you to look at.

- `services/cms/mockApi.js` should have a `fetchAdvanceSettings` function so this works in gallery view.

- Currently, `useArrowNav` which lets you move up and down selections by pressing the Up and Down arrow buttons will trigger an error. This can be used for `AdvancedTypeSelect` as well by accounting for `AdvanceProblems` in `useArrowNav` and adding `next`, `prev` properties on the `AdvanceProblems`. I could code this in if you like. I should have made it more error-proof.

src/editors/containers/ProblemEditor/components/SelectTypeModal/hooks.js
KristinAoki: I don't think that creating another reducer makes the call more readable and direct. This is the only place the reducer would be used so it seems unnecessary to create a new reducer function for just one call. Unlike `setProblemType` which is probably used in multiple location.

rayzhou-bit: Ok, that makes sense to me. I was thinking about it from the point of view of reading an onSelect call and getting an understanding of what it's doing based on the inputs.

src/editors/containers/ProblemEditor/components/SelectTypeModal/hooks.js
```diff
@@ -28,17 +32,25 @@ export const useArrowNav = (selected, setSelected) => {
...
+    const advanceTypeValues = Object.values(AdvanceProblemKeys);
     switch (e.key) {
       case 'ArrowUp':
-        if (ProblemTypes[selected].prev) {
+        if (problemTypeValues.includes(selected) && ProblemTypes[selected].prev) {
           setSelected(ProblemTypes[selected].prev);
           document.getElementById(ProblemTypes[selected].prev).focus();
+        } else if (advanceTypeValues.includes(selected) && AdvanceProblems[selected].prev) {
+          setSelected(AdvanceProblems[selected].prev);
+          document.getElementById(AdvanceProblems[selected].prev).focus();
```
rayzhou-bit, line 44: I just realized that Form.Radio has the arrow functionality coded in... I'm so sorry I did not realize that earlier. You could remove the `else if` with advanced types and arrow navigation will still work.

The type checking looks great.

src/editors/data/constants/problem.js
```diff
@@ -27,6 +27,7 @@ export const ProblemTypes = StrictDict({
     preview: singleSelect,
     description: 'Specify one correct answer from a list of possible options',
     helpLink: 'https://edx.readthedocs.io/projects/edx-partner-course-staff/en/latest/exercises_tools/multiple_choice.html',
+    prev: ProblemTypeKeys.TEXTINPUT,
```
rayzhou-bit, line 30: Good catch. That will match how `Form.Radio` works.

rayzhou-bit (approved):

PR #176, opened 2022-12-22
feat: add import transcripts from youtube

KristinAoki:
JIRA Ticket: [TNL-9954](https://2u-internal.atlassian.net/browse/TNL-9954)

This PR adds a card when transcripts from youTube have the ability to be imported.


src/editors/data/services/cms/api.js
```diff
@@ -51,6 +51,33 @@ export const apiMethods = {
...
+    youTubeId,
+  }) => {
+    const getJSON = `{"locator":"${blockId}","videos":[{"mode":"youtube","video":"${youTubeId}","type":"youtube"}]}`;
+    return get(
+      urls.replaceTranscript({
+        studioEndpointUrl,
+        parameters: encodeURIComponent(getJSON),
+      }),
+    );
+  },
```
rayzhou-bit, line 80: Could you add these to `mockApi.js` as well for gallery view to work?

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/TranscriptWidget/ImportTranscriptCard.jsx
```diff
@@ -0,0 +1,60 @@
...
+  importTranscript,
+}) => (
+  <Stack gap={3} className="border rounded border-primary-200 p-4">
+    <ActionRow className="h5">
+      <FormattedMessage {...messages.importHeader} />
+      <ActionRow.Spacer />
+      <IconButton
+        src={Close}
+        iconAs={Icon}
+        onClick={() => setOpen(false)}
```
rayzhou-bit, line 30: I think you can write a test for this similar to the behavior inspection in the EditorContainer test file.

rayzhou-bit (changes requested):
lgtm! Just need that one test

src/editors/data/services/cms/api.js
```diff
@@ -51,6 +51,33 @@ export const apiMethods = {
       data,
     );
   },
+  checkTranscripts: ({
```
connorhaugh, line 54: "check" doesn't seem super descriptive as to what is going on here. is there any information other than the youtube transcript check which transpires at this endpoint?

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/TranscriptWidget/index.jsx
```diff
@@ -186,6 +194,7 @@ export const mapStateToProps = (state) => ({
   transcripts: selectors.video.transcripts(state),
   allowTranscriptDownloads: selectors.video.allowTranscriptDownloads(state),
   showTranscriptByDefault: selectors.video.showTranscriptByDefault(state),
+  transcriptCheckCommand: selectors.video.transcriptCheckCommand(state),
```
connorhaugh, line 197: instead of using a variable stored in the video state here as a variable, it would be simpler to just query the request state of the transcriptcheck call ('selectors.requests.isCompleted(transcriptCheck)' or something like that). It will reduce the complexity of the video redux state.

src/editors/data/services/cms/api.js
KristinAoki: This API is checking if the video has any transcripts available for import.

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/TranscriptWidget/index.jsx
KristinAoki: I need the selector because that has a bool that determines if there are transcripts available to import. So I am not relying on the completion of the API call to determine whether to show the card or not.

connorhaugh: ah gotcha -- late afternoon brain IDk why it got into my head that that was the feature lol

connorhaugh (approved):
LGTM

PR #178, opened 2023-01-04
feat: video editor fix to disappearing transcript

rayzhou-bit:
https://2u-internal.atlassian.net/browse/TNL-10300

Issue: Currently, when user updates the video id or video url field, the rest of the settings will get reset. This occurs onBlur, which means if the user updates one of these fields and immediately clicks Save, they will not realize the form has completely updated. Another point of confusion is that there is no notification that the settings are reset. 
Users filed this issue because importing youtube transcripts reset would set a video id, which would then reset the youtube video url.

Solution: Do not reset settings when video id or video url field is updated. This copies legacy behavior and means both id and url can be saved. User work is saved where possible

Testing:
1. Create a new video block with a video url that has transcripts (can use this one https://youtu.be/B-EFayAA5_0).
2. Import transcripts.
3. Studio page should not show "No playable video sources found" on that block.
4. Opening up the block again should show both a video id and a video url.

Minor Refactor Note: Biggest part of this was removing react-thunk as we are mostly not using it. We were using useSelector hook to pull form store anyways.


connorhaugh:
Are there testing instructions or an exact description of the solution you are creating? The ticket seems to insinuate this is a browser issue, but you are re-organizing a lot of code as well it seems.

package.json
```diff
@@ -66,6 +66,7 @@
     "canvas": "^2.10.1",
     "codemirror": "^6.0.0",
     "fast-xml-parser": "^4.0.10",
+    "imagemin-optipng": "^8.0.0",
```
connorhaugh, line 69: why are you adding this package if you never use it?

connorhaugh: If we are furthermore, is this a quality package we can trust? last release came out 3 years ago

kenclary (commented):
looks like three new lines to get new tests

package.json
rayzhou-bit: Good catch, I didn't mean to have this in the PR. This was part of devstack debugging.

connorhaugh (commented):
>  biggest refactor was to remove react-thunk as we are mostly not using it. We were using useSelector hook to pull form store anyways.

there are no changes to any of our Thunk Action files???

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/VideoSourceWidget/index.jsx
```diff
@@ -144,12 +143,6 @@ export const VideoSourceWidget = ({
 VideoSourceWidget.propTypes = {
   // injected
   intl: intlShape.isRequired,
-  // redux
-  updateField: PropTypes.func.isRequired,
 };
-export const mapStateToProps = () => ({});
 
-export const mapDispatchToProps = (dispatch) => ({
```
connorhaugh, line 152: Do you mean making this component no longer "connected?"

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/VideoSourceWidget/hooks.jsx
```diff
@@ -1,70 +1,27 @@
...
+  updateVideoURL: ({ e }) => dispatch(actions.video.updateField({ videoSource: e.target.value, videoId: '' })),
+  updateVideoId: ({ e }) => dispatch(actions.video.updateField({ videoId: e.target.value, videoSource: '' })),
+});
+
+export const fallbackHooks = ({ fallbackVideos, dispatch }) => ({
+  addFallbackVideo: () => dispatch(actions.video.updateField({ fallbackVideos: [...fallbackVideos, ''] })),
+  deleteFallbackVideo: (videoUrl) => {
+    const updatedFallbackVideos = [];
+    let firstOccurence = true;
+    fallbackVideos.forEach(item => {
```
connorhaugh, line 13:
```suggestion
    const updatedFallbackVideos = fallbackVideos.splice(fallbackVideos.indexOf(videoUrl),1)
```
Does the same thing, is cleaner.

connorhaugh (commented):
Just in general, a few notes but seems you need to add a smidgen of test coverage.

jristau1984:
> Just in general, a few notes but seems you need to add a smidgen of test coverage.

Also, a note from yesterday's standup to add tests that help document the intricacies of this behavior.

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/VideoSourceWidget/index.jsx
rayzhou-bit: Yes, from what I understand, you shouldn't need to connect via redux-thunk when you use the useSelector hook.

rayzhou-bit:
> > biggest refactor was to remove redux-thunk as we are mostly not using it. We were using useSelector hook to pull form store anyways.
> 
> there are no changes to any of our Thunk Action files???

Thunk Actions and redux-thunk are not the same things. redux-thunk is used to set up redux state and you can see this in data/store.js. Our thunk actions are an in-between processing to convert the raw values we fetch from our api and put them neatly in our redux store.

Ben has been working on removing redux-thunk and going full hook: https://github.com/openedx/frontend-app-learner-dashboard/pull/98/files

KristinAoki (commented):
Everything works as expected. Is there a reason that the `package-lock.json` files are included in this PR?

rayzhou-bit:
> Everything works as expected. Is there a reason that the `package-lock.json` files are included in this PR?

I should get those out of the PR. Thanks for catching that!

rayzhou-bit:
> Everything works as expected. Is there a reason that the `package-lock.json` files are included in this PR?

I couldn't get the package-lock changes to disappear. But I think it should be fine since I checked out the package-lock files from an updated origin/main.

PR #181, opened 2023-01-09
feat: unselect multiple choices type multi->single

KristinAoki:
JIRA Ticket: [TNL-10286](https://2u-internal.atlassian.net/browse/TNL-10286)

This ticket resolves the bug that allowed multiple answers to be selected when switching between multi-select problem types and single select problem types. Now when a user switches from a multi-select problem to a single select problem with multiple selected answers, all the answers will be deselected. If only one answer is selected, the answer is persisted.


rayzhou-bit (approved):
lgtm! I tested it out on my local. You just need to get the code coverage up on the last few lines.

PR #184, opened 2023-01-11
feat: allow feedback to copy when problem changed

KristinAoki:
JIRA Ticket: [TNL-10285](https://2u-internal.atlassian.net/browse/TNL-10285)

This PR adds conditions that allow the "is selected" feedback field for multi select to persist to the single feedback field when the user switched to another problem type. The added conditions also update the multi select "is selected" feedback field when switching to multi select from another problem type.


rayzhou-bit (commented):
Are there any reasons for `feedback` and `selectedFeedback` to be different? Could we just remove `selectedFeedback`?

KristinAoki:
> Are there any reasons for `feedback` and `selectedFeedback` to be different? Could we just remove `selectedFeedback`?

I originally tried that, but when I was saving I received an error message about the formatting. However, after my current fix, I realized the two were unrelated. I will make your suggested change.

rayzhou-bit (commented):
Just need some more unit testing and this looks good to go!

rayzhou-bit (approved):

PR #187, opened 2023-01-11
fix: match new frontend behavior with legacy behavior (zero problem attempts is zero attempts, null attempts is infinite); negative values disallowed and forced to zero. TNL-10324.

kenclary:


src/editors/containers/ProblemEditor/components/EditProblemView/SettingsWidget/hooks.test.js
```diff
@@ -178,7 +178,19 @@ describe('Problem settings hooks', () => {
       const value = 0;
       output.handleMaxAttemptChange({ target: { value } });
       expect(updateSettings)
-        .toHaveBeenCalledWith({ scoring: { ...scoring, attempts: { number: value, unlimited: true } } });
+        .toHaveBeenCalledWith({ scoring: { ...scoring, attempts: { number: value, unlimited: false } } });
```
jesperhodge, line 181: Part of the issue ACs was
"- “0” attempts will disable the problem from having submissions
- NULL attempts will allow for infinite/unlimited submissions"

Maybe it would make sense to add these as test cases to make sure we have (and keep) the correct behavior now?

jesperhodge (changes requested):
Looks good, I just added a small suggestion about the tests.

src/editors/containers/ProblemEditor/components/EditProblemView/SettingsWidget/hooks.test.js
kenclary: I've updated the tests to test for an actual null value (and other edge cases), rather than just a representative non-numeric value.

Also note, the actual behavior of disabling submissions with 0 attempts, or allowing infinite attempts if `unlimited` is set to true, is handled by another frontend entirely. We just send the values when we save.

jesperhodge: Great, that looks good!

jesperhodge (approved):

PR #189, opened 2023-01-11
fix: change # destination to advance settings url

KristinAoki:
JIRA Ticket: [TNL-10326](https://2u-internal.atlassian.net/browse/TNL-10326)

This PR sets the "Set a default value in advanced settings" link to `<studiourl>/settings/advanced/<courseID>`. Previously the link was set to `#` so it was opening the problem editor page, but in a different tab.

### DIFF
--- a/src/editors/containers/ProblemEditor/components/EditProblemView/SettingsWidget/settingsComponents/ShowAnswerCard.jsx
+++ b/src/editors/containers/ProblemEditor/components/EditProblemView/SettingsWidget/settingsComponents/ShowAnswerCard.jsx
@@ -1,9 +1,11 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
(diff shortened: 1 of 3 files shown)


src/editors/containers/ProblemEditor/components/EditProblemView/SettingsWidget/settingsComponents/ShowAnswerCard.jsx
```diff
@@ -1,9 +1,11 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
```
rayzhou-bit, line 3: This is something we should discuss more as a group first but I think we should move away from using connect and move towards using the useSelector hook. The studioEndpointUrl could be pulled from state through useSelector(selectors.app.studioEndpointUrl).

rayzhou-bit (approved):
lgtm!

PR #190, opened 2023-01-12
feat: video duration settings refactor

rayzhou-bit:
This PR aims to refactor a portion of the video settings code with the following goals:

1. Compartmentalization of the duration widget into its own folder. Component rendering and logic code are found solely under the DurationWidget folder rather than across multiple files in the main video settings component folder.
2. Removal of react-thunk usage. We should move towards using the useSelector hook to access the redux library rather than `connect`ing with `mapStateToProps` and `mapDispatchToProps`. This is mentioned as part of react-redux best practice: https://redux.js.org/style-guide/#use-the-react-redux-hooks-api. Connect still works and will continue to be supported, but useSelector is suggested.
3. Increase readability by ending calls to `handlers.js` and `hooks.js` in the components folder. The functions in these files call each other multiple times. This is hard to keep track of when trying to figure out what the code does.


kenclary (commented):
codecov calls out a couple lines that aren't tested (basically things like max time). We should test for that if we don't already. Otherwise, this looks good, and I'm psyched to see it happen.

connorhaugh:
What's your justification for #2 ?

jristau1984:
> What's your justification for `#2` ?

Expanding on this... if we are making a decision to change, that decision may need to persisted into the ADRs for the repo as well.

rayzhou-bit:
Both useSelector and connect can be used, but react-redux strongly recommends the hooks api. https://redux.js.org/style-guide/#use-the-react-redux-hooks-api

(I updated description with this info.)

rayzhou-bit:
After some discussions on useSelector vs connect, I decided to research how useSelector re-rendering worked with slices of state. 

From https://redux.js.org/usage/deriving-data-selectors:
"Selectors used with useSelector or mapState will be re-run after every dispatched action, regardless of what section of the Redux root state was actually updated. Re-running expensive calculations when the input state sections didn't change is a waste of CPU time, and it's very likely that the inputs won't have changed most of the time anyway."

So, although there is more boilerplating, connect will re-render less and give us better performance. I'll be updating my PR to use connect instead of useSelector.

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/DurationWidget/hooks.js
```diff
@@ -0,0 +1,219 @@
+import { useEffect, useState } from 'react';
+import { useSelector } from 'react-redux';
+
+import { actions, selectors } from '../../../../../../data/redux';
+import messages from '../messages';
+
+import * as module from './hooks';
```
jesperhodge, line 7: Circular dependencies can be dangerous and lead to bugs, at least they make the code very complicated. There is no need for a pattern like this, you can just declare any functions you want to use in this file in the top as `const myFunction` and use them directly and then export them below.

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/DurationWidget/hooks.js
```diff
@@ -0,0 +1,219 @@
...
+
+const durationMatcher = /^(\d{0,2}):?(\d{0,2})?:?(\d{0,2})?$/i;
+
+export const durationWidget = ({ dispatch }) => {
+  const reduxStartStopTimes = useSelector(selectors.video.duration);
+  const setReduxStartStopTimes = (val) => dispatch(actions.video.updateField({ duration: val }));
+  const initialState = module.durationString(reduxStartStopTimes);
```
jesperhodge, line 14: There is no need for this, as the `durationString` function is in the same file. You can just move the function to the top as `const durationString = (duration) => ({` and export it at the bottom. Here, you can call `setUnsavedStartStopTimes(durationString(reduxStartStopTimes` instead.

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/DurationWidget/hooks.js
```diff
@@ -0,0 +1,219 @@
...
+
+  useEffect(() => {
+    setUnsavedStartStopTimes(module.durationString(reduxStartStopTimes));
+  }, [reduxStartStopTimes]);
+
+  return {
+    reduxStartStopTimes,
+    unsavedStartStopTimes,
+    onBlur: (index) => (
+      (e) => module.updateDuration({
```
jesperhodge, line 25: See other comment about module antipattern

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/DurationWidget/hooks.js
```diff
@@ -0,0 +1,219 @@
...
+        reduxStartStopTimes,
+        setReduxStartStopTimes,
+        unsavedStartStopTimes,
+        setUnsavedStartStopTimes,
+        index,
+        durationString: e.target.value,
+      })
+    ),
+    onChange: (index) => (
+      (e) => setUnsavedStartStopTimes(module.onDurationChange(unsavedStartStopTimes, index, e.target.value))
```
jesperhodge, line 35: See other comment about module antipattern

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/DurationWidget/hooks.js
```diff
@@ -0,0 +1,219 @@
...
+    ),
+    onKeyDown: (index) => (
+      (e) => setUnsavedStartStopTimes(module.onDurationKeyDown(unsavedStartStopTimes, index, e))
```
jesperhodge, line 38: See other comment about module antipattern

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/DurationWidget/hooks.js
```diff
@@ -0,0 +1,219 @@
...
+    ),
+    getTotalLabel: ({ duration, subtitle, intl }) => {
+      if (!duration.stopTime) {
+        if (!duration.startTime) {
+          return intl.formatMessage(messages.fullVideoLength);
+        }
+        if (subtitle) {
+          return intl.formatMessage(
+            messages.startsAt,
+            { startTime: module.durationStringFromValue(duration.startTime) },
```
jesperhodge, line 48: See other comment about module antipattern

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/DurationWidget/hooks.js
```diff
@@ -0,0 +1,219 @@
...
+          );
+        }
+        return null;
+      }
+      const total = duration.stopTime - (duration.startTime || 0);
+      return intl.formatMessage(messages.total, { total: module.durationStringFromValue(total) });
```
jesperhodge, line 54: See other comment about module antipattern

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/DurationWidget/hooks.js
```diff
@@ -0,0 +1,219 @@
...
+export const updateDuration = ({
+  reduxStartStopTimes,
+  unsavedStartStopTimes,
+  setReduxStartStopTimes,
+  setUnsavedStartStopTimes,
+  index,
+  inputString,
+}) => {
+  let newDurationString = inputString;
+  let newValue = module.valueFromDuration(newDurationString);
```
jesperhodge, line 110: See other comment about module antipattern

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/DurationWidget/hooks.js
```diff
@@ -0,0 +1,219 @@
...
+  // stopTime must be at least 1 second after startTime, except 0 means no custom stopTime
+  if (index === 'stopTime' && newValue > 0 && newValue < (reduxStartStopTimes.startTime + 1000)) {
+    newValue = reduxStartStopTimes.startTime + 1000;
+  }
+  // startTime must be at least 1 second before stopTime, except when stopTime is less than a second
+  // (stopTime should only be less than a second if it's zero, but we're being paranoid)
+  if (index === 'startTime' && reduxStartStopTimes.stopTime >= 1000 && newValue > (reduxStartStopTimes.stopTime - 1000)) {
+    newValue = reduxStartStopTimes.stopTime - 1000;
+  }
+  newDurationString = module.durationStringFromValue(newValue);
```
jesperhodge, line 128: See other comment about module antipattern

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/DurationWidget/hooks.js
```diff
@@ -0,0 +1,219 @@
...
+  };
+};
+
+/**
+ * durationString(duration)
+ * Returns the display value for embedded start and stop times
+ * @param {object} duration - object containing startTime and stopTime millisecond values
+ * @return {object} - start and stop time from incoming object mapped to duration strings.
+ */
+export const durationString = (duration) => ({
```
jesperhodge, line 65: This function, since it doesn't get hoisted, is better placed higher up in the file so it can be used without needing the "module.myFunction" call. Same for `durationStringFromValue`.

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/DurationWidget/hooks.js
```diff
@@ -0,0 +1,219 @@
...
+import { useSelector } from 'react-redux';
+
+import { actions, selectors } from '../../../../../../data/redux';
+import messages from '../messages';
+
+import * as module from './hooks';
+
+const durationMatcher = /^(\d{0,2}):?(\d{0,2})?:?(\d{0,2})?$/i;
+
+export const durationWidget = ({ dispatch }) => {
```
jesperhodge, line 11: The name is a bit ambiguous because now you have a component and a function with the same name. Also, this function calls other hooks inside it (like `useSelector`), which makes it a custom hook. That is why the name must start with `use`. (For example, `useDuration` would probably be a good name). I know we don't name custom hooks that way so far in the project, but that is a mistake - have a look at the React docs here (https://reactjs.org/docs/hooks-custom.html) if you want to know more. Since it is important to distinguish between custom hooks and other functions, React makes it very clear that it is mandatory to name custom hooks with `use`. If we don't do that, it will be confusing and make other React developers have problems understanding our code.

package-lock.json
```diff
@@ -192,15 +192,6 @@
         "node": ">=6.9.0"
```
jesperhodge, line 192: If you haven't changed anything in the package.json file, is there a reason the package.lock needs to change? Otherwise it should not be included in this PR because there is always a danger of breaking production dependencies.
A tip: if you want to install dependencies from a branch but don't want to change your package.lock, you can use `npm ci` instead.

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/DurationWidget/hooks.test.js
```diff
@@ -0,0 +1,328 @@
...
+        duration: { startTime: '00:00:00' },
+        index: 'startTime',
+        event: 'eVeNt',
+      };
+      hook = hooks.onDurationKeyDown;
+    });
+    it('enter event: calls blur()', () => {
+      props.event = { key: 'Enter' };
+      const blurSpy = jest.spyOn(document.activeElement, 'blur');
+      hook(props.duration, props.index, props.event);
```
jesperhodge, line 303: Not sure about the naming here - from the name, it is pretty unclear what a method `hook(...)` is supposed to do. Create a new react hook?

jesperhodge (changes requested):
Hey, I really like the refactoring you did here. I left some comments for further improvements, but all in all great job.

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/DurationWidget/hooks.js
rayzhou-bit: I've changed the code to use connect instead of useSelector now.

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/DurationWidget/hooks.js
rayzhou-bit: We discussed this on slack but I'll layout the result of that discussion here. Refactoring out self-import should be something we take a look at in the future, but for now, I'll move ahead on this PR with self-import in play.

package-lock.json
rayzhou-bit: I did not mean to change these. I think my machine has been evaluating `npm install` differently somehow.

jesperhodge (changes requested):
There's an error when you get to video settings, click on duration -> start time input, and then on stop time input

"Cannot read properties of undefined (reading 'trim')"

jesperhodge (approved):
Awesome!

PR #194, opened 2023-01-12
Fix: advanced problem template reference

connorhaugh:
As I was writing the select type egress, I fixed a bug where the wrong rawolx object was being loaded in for basic problem types. this fixes it for advanced ones as well.


KristinAoki (approved):
A follow up ticket about the ui will need to be made for the footer when in the raw editor.

PR #206, opened 2023-01-20
feat: remove drag and drop from advanced problems

KristinAoki:
JIRA Ticket: [TNL-10380](https://2u-internal.atlassian.net/browse/TNL-10380)

This PR removes the deprecated drag and drop advanced problem type from the list of advanced problem types. Drag and drop V2 is still available in studio via the "Advanced" component.


connorhaugh (approved):
Looks good to me, you might want to mention in the commit message that the v2 drag and drop problem will still be exposed in the "advanced" section in studio?

PR #211, opened 2023-01-24
feat: add function to remove empty hints

KristinAoki:
JIRA Ticket: [TNL-10384](https://2u-internal.atlassian.net/browse/TNL-10384)

This PR adds an `onBlur` function to the Hint Row that deletes the hint if it is empty. Currently when a hint is added focus remains on the "Add Hint' button. Therefore, if a user never clicks on the hint input field, the row will not be deleted unless the user clicks the delete button. This was not requested in the ticket, but might be something to consider later.


src/editors/containers/ProblemEditor/components/EditProblemView/SettingsWidget/settingsComponents/HintRow.jsx
```diff
@@ -1,48 +1,43 @@
...
-        </Form.Group>
-      </Col>
-
-      <Col xs={2}>
-        <IconButton
-          src={Delete}
-          iconAs={Icon}
-          alt={intl.formatMessage(messages.settingsDeleteIconAltText)}
-          onClick={handleDelete}
-          variant="secondary"
```
kenclary, line 35: are we sure we don't want `variant="secondary"` in the new version?

KristinAoki: yes

kenclary (approved):

PR #219, opened 2023-01-27
fix: question field uses real placeholder instead of template. fixed blank question handling. TNL-10395, TNL-10411.

kenclary:
fix: question field uses real placeholder instead of template. fixed blank question handling. TNL-10395, TNL-10411.


KristinAoki (approved):
LGTM

PR #220, opened 2023-01-27
docs: reducing complexity

jesperhodge:
- docs: add adr
- docs: update adrs


docs/decisions/0006-general-principles-for-react-development.md
```diff
@@ -0,0 +1,29 @@
...
+
+Reason: The community has spent years learning and improving upon what works well and what doesn't, and this has led to the patterns described there. The docs are in general not opinionated and violating their principles has caused applications to have many bugs. When we use established principles, we make it easy for people to understand the code, especially if they are newcomers to the project or community members.
+
+IV. We follow the KISS principle - "Keep it simple, stupid", in order to make it easy for any developer to understand the code.
+
+Reason: We work agile and with an open source community. A common rule of thumb is that a developer with just 3 months of experience should be able to understand code. We should enable developers to work on this codebase without spending a lot of time trying to understand what is going on.
+
+V. We do not increase code complexity in order to make it easier to write tests, unless absolutely necessary and for good reasons. (Reducing complexity to improve testability is fine.)
+
```
bszabo, line 26: Yes and no. Using dependency injection to make it easy to mock out dependencies is essential for testability. One might argue that the inversion of control that comes with that is non-intuitive and makes the code harder to understand. But it's a really sound design and testability principle. We're in agreement if dependency injection is not viewed as increasing code complexity.

bszabo: You mentioned the KISS design principle above, but we should equally be adhering to other design principles such as the single-responsibility principle. For example, mixing UI and business logic may be intuitive when first implementing a feature, but may well make the testing harder and may actually be harder to use in the future, as requirements change. This is probably addressed in the open source community practices that you've referenced, but I worry about a blanket statement that we shouldn't structure our code for testability.

Put another way, following the best design principles is good engineering and is good for testability, but it may not be ideal for clarity to someone with three months' experience. I believe a more appropriate metric is that the code is easily accessible to someone with limited domain knowledge, but who has experience following good design principles.

docs/decisions/0007-reducing-complexity.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Motivation
+
+High complexity is in general a concerning problem: It can lead to untested legacy code, a high amount of bugs, bad maintainability, slow development, and so on. These decisions are concrete changes we want to make to reduce complexity.
+
+## Decisions
+
+### Hooks
+
+Decision: Hooks always start with `use` (for example `useCounter`). This applies to custom hooks. A custom hook is any function that calls other hooks, especially state and lifecycle hooks like `useEffect` or `useState`. Functions that do not call other hooks never start with `use`.
```
bszabo, line 11: Is this a design decision or a coding convention? Sounds more like the latter.

docs/decisions/0007-reducing-complexity.md
```diff
@@ -0,0 +1,39 @@
...
+
+Reason: See general principles (3). See [React Docs](https://beta.reactjs.org/learn/reusing-logic-with-custom-hooks#hook-names-always-start-with-use)
+
+### Contents of hooks.js
+
+Decision: A hooks.js file lives in the directory of a React component, and must contain only custom hooks. All other functions that are not custom hooks must be in other files. (You may make an exception for functions that you consider private to that hooks.js file and do not expect to be used anywhere else.) All custom hooks that belong to the parent component must be in that hooks.js file. This does not apply for hooks that are intended to be reusable.
```
bszabo, line 17: Ditto on design vs. implementation. You're bringing up good points that will improve the quality of our code, and they belong somewhere. I'm just not sure that this is the place. Perhaps the team needs a coding style document, and these go there.

docs/decisions/0007-reducing-complexity.md
```diff
@@ -0,0 +1,39 @@
...
+
+Reason: It is common practice to have a "hooks.js" file and you will only expect hooks in there. We put all (except for reused) hooks in there in order to make it easy to find hooks and distinguish between them and normal functions.
+
+### Reusable hooks
+
+Decision: Hooks intended to be reusable live in shared directories. Either the directory name or the filename they are in must signify that they are hooks.
```
bszabo, line 23: Design or coding question?

docs/decisions/0007-reducing-complexity.md
```diff
@@ -0,0 +1,39 @@
...
+
+Reason: Easily distinguish hooks from other functions
+
+### Self-imports
+
+Decision: We avoid self-imports and replace them wherever we find them.
+
+Reason: See general principles (5). They have unnecessary complexity. The conceptual idea behind it is not so clear: the way we program in JS is usually a one-directional flow where we declare a const for example, either import it or declare it in the same file, then use it down the line and export it. Conceptually the idea that you declare a constant, export it, then import it into the same file, then export again makes everything pretty convoluted, and we also have extra syntactical overhead for it. The code itself does not explain why the logic is so complicated and what it is there for - because it is only there for the tests. From a code perspective it is not logical to use such a logic. We do not need self-imports for writing tests, other projects generally do not use them.
```
bszabo, line 31: Is the broader issue here that good stewardship calls for all test-related logic to live in test files, and for production files to only have code that contributes to features? And that our use of re-importing is causing test logic to live in production files? I haven't encountered this pattern, but wonder whether it's actually a "design for testability" issue. What is it that the re-importing is enabling for tests? Is there an alternate design that keeps test logic in test files?

docs/decisions/0007-reducing-complexity.md
```diff
@@ -0,0 +1,39 @@
...
+
+### component state
+
+Decision: We use useState in the way it is demonstrated in the React docs, and we do not export a constant `state` from our hooks.js files.
+
+Reason: See general principles (3, 5). React has developed an elegant, well-working and testable state hook, and we do not need to abstract that with unnecessary complexity. This is also not needed for testability. Mocking useState is very difficult because you are not intended to. If you really need to do so, you can still absolutely do it, and this is a testing issue and should not affect the code in any way. However, it is not at all recommended to test state directly. Testing state directly, e.g. by mocking useState, will lead to false positives and false negatives in the tests, and as of React 18 useState behavior is not deterministic in development (&test?) environments. There is also the possibility of working with `useReducer`, (React local state hook, not redux reducer), which should be a bit easier to test.
```
bszabo, line 39: Are you suggesting that configuring a test with a specific starting state ought to be done with the mutator returned by useState, and never by mocking useState itself? Or are you suggesting that a test should never be in the business of directly setting state as part of its setup? (This is a clarification question only).

docs/decisions/0006-general-principles-for-react-development.md
jesperhodge: I fully agree with these points. It would be great to find a better wording for this. My motivation for the above statement V. is that I see some unusual patterns in our codebase that I haven't seen used much in React codebases, which I consider bad / unnecessarily complex patterns, and they have been explicitly justified in an earlier ADR with saying that improved testability should justify higher complexity. I am attempting to understand the underlying problem, and it seems to me to have something to do with believing there is a tradeoff - that for higher testability you need to increase code complexity. My experience is usually that higher testability best goes hand in hand with a really good code design. I would also add to your first point that dependency injection as you say is "a really sound design" principle as well, which justifies it. We don't really have the option to do this easily in React, but I want to make the point that we should improve testability **along with** good design and not **instead of** good design.
Do you have any ideas how to frame this point in a way that is better and less of a blanket statement?

docs/decisions/0007-reducing-complexity.md
jesperhodge: You could call it a coding convention, but it is more of a rule. We use React and React is very good in not being opiniated. However, React docs state custom hooks **MUST** start with "use" and this is not optional but mandatory. They give very good reasons for it (the link contains it all) and if we don't follow that we don't just create bad code design, but confuse other React developers who look at the code base.

docs/decisions/0007-reducing-complexity.md
jesperhodge: Yeah, I don't have a preference where these points should go.

docs/decisions/0007-reducing-complexity.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Decisions
+
+### Hooks
+
+Decision: Hooks always start with `use` (for example `useCounter`). This applies to custom hooks. A custom hook is any function that calls other hooks, especially state and lifecycle hooks like `useEffect` or `useState`. Functions that do not call other hooks never start with `use`.
+
+Reason: See general principles (3). See [React Docs](https://beta.reactjs.org/learn/reusing-logic-with-custom-hooks#hook-names-always-start-with-use)
+
+### Contents of hooks.js
```
jesperhodge, line 15: A note on this: there are two parts. A. a hooks.js file should only contain hooks - I consider important because it is a convention in the React world; and B. anything else I wrote about organizing that file and were to put hooks and functions - is just an idea how we could structure it, we can change that anyway you feel like.

docs/decisions/0007-reducing-complexity.md
jesperhodge: Just for throwing out ideas where to put stuff if we don't have all our functions in the hooks file. Maybe we should not have this in here as it's not an important point?

docs/decisions/0007-reducing-complexity.md
jesperhodge: Yeah there are a lot of reasons that have apparently led up to this happening. The main reason for these self-imports in our project was that if you have two functions A and B in the same file and you want to test one and mock the other out you have some problems in JS. But I have never seen this as an obstacle in other codebases, and it is so easy to just throw one of the functions into another file, which is fine from a design standpoint - considering that webpack will merge those two files together anyway, so it has no performance implications to have a bunch of files.

docs/decisions/0007-reducing-complexity.md
jesperhodge: Yeah it's actually a whole topic and very interesting point, and there are many very interesting articles on the topic. I would love to talk about the way we test in React, but I think that is very out of scope here and a larger topic and more controversial, so I just added this note to explain why you can get around without needing this weird pattern. Generally the idea out there is that you don't test your React hooks and state directly because they are not pure functions. They are all about side effects that control the lifecycle of the React component. A state hook behaves differently in development, test and production environments starting from React 18, so it is not deterministic. The hook calling setState may work amazingly well but from calling it you cannot see what it does with the lifecycle of the component, so you often have false positives. You also may have false negatives because the lifecycle and state may be controlled by different elements that make it work nonetheless. Basically what we are testing are not state hooks, but the component state and lifecycle, and we need to make that behave like in reality. So what the recommended way of testing React state is is to test the whole component as your unit and test the behavior when the component renders, when it loads data, when a button is clicked and so on. So now what if you have a React hook that is built for many components to reuse? And we don't have many of those anyway. If the hook is just part of your one component, best to test it as part of that component. Otherwise, you would in theory build a mock/testing component and add the hook to that component; but react-testing-library has tools to do that for you so you don't have to worry about it actually.

arbrandes:
@jesperhodge, there are ADRs in open-edx-proposals as well; usually ones that go into dirtier details than in the OEP itself.  Take [OEP-45](https://github.com/openedx/open-edx-proposals/blob/master/oeps/architectural-decisions/oep-0045-arch-ops-and-config.rst), for instance, and [its ADRS](https://github.com/openedx/open-edx-proposals/tree/master/oeps/architectural-decisions/oep-0045/decisions) (there's [a new one](https://github.com/openedx/open-edx-proposals/pull/372) coming down the pipes).

So, for instance, we can have a [best-practices](https://github.com/openedx/open-edx-proposals/tree/master/oeps/best-practices) OEP with the general ideas you put forward here, then a more specific ADR for React.

jesperhodge:
I made a new commit with the following purpose: in the reducing-complexity ADR, I removed anything that is just preference or convention, like where to put hook files. What remain should only be steps that are justified and necessary for fixing concrete problems.

I will also remove the "General Principles" ADR to later put it somewhere else - probably have it as an OEP, like Adolfo suggested. It needs to be heavily changed and iterated on, anyway.

The remaining ADR on reducing-complexity, only solving certain concrete problems in this codebase, I think is ready to change from draft to normal PR. There are more points we can find to improve the code design of this project, maybe add to that ADR, but they can be added in a later PR.

Points we have decided upon when this ADR is approved and merged that are in conflict with what ADR-0005 says, we should probably remove from 0005 when this is done.

docs/decisions/0007-reducing-complexity.md
jesperhodge: After some time to think: It doesn't matter much how you call it. What's relevant here is that you must be able to distinguish hooks from functions, and this is the only correct way to do it as decided by the React team. I can recommend reading more about it in the beta docs, it is enlightening.

docs/decisions/0007-reducing-complexity.md
jesperhodge: I think these points should go here as long as they pertain just concretely to the project (see Adolfo's suggestion below, I am in favor of adding a more general doc or ADR for React best practices. We also have an internal wiki document on React best practices that is basically empty right now.) But I removed any design decisions here except for the point of not having non-hooks in hooks files, and the reason for that is that it is based on community standards and not on a subjective design decision.

docs/decisions/0007-reducing-complexity.md
jesperhodge: I removed this point

jesperhodge:
We have decided to close this PR. The idea after discussing it with the team is that this is not something that belongs into an ADR, exactly. We have discussed that most of these discussions tie into testing in some way; and that we would rather like to discuss improving our testing strategy for this repository. These code quality improvements may result organically from improving our tests. On top of that, we have an open issue for improving folder structure, created by Connor. So basically we want to tackle the larger challenges in the repo first and go from there.

PR #221, opened 2023-01-30
feat: add label button to question tinyMCE

KristinAoki:
JIRA Ticket: [TNL-10396](https://2u-internal.atlassian.net/browse/TNL-10396)

This PR adds a custom "Label" button to the question widget's tinyMCE. Label tags are used by screen readers and emphasis the main question. Previously label tags were converted to strong tags, and with this change the tags are maintained. 

One thing to note is that tinyMCE requires all text to be wrapped in valid block tags. Therefore, label tags have to be wrapped inside paragraph tags. To keep the rendering of the olx inside the tinyMCE editor from rearranging the text, all labels have to be in their own paragraph. This means that whenever a label is added, a new paragraph is also added.

Simple Example:
Raw OLX
```
<p>This is an example</p><label>This is the important part</label>
```
translates to
```
<p>This is an example</p><p><label>This is the important part</label></p>
```
Label inside paragraph example:
Raw OLX
```
<p>This is an example with<label>a label inside a paragraph</label></p>
```
translates to
```
<p>This is an example with</p><p><label>a label inside a paragraph</label></p>
```
It is important to note that this current implementation assumes that the labels will be wrapped in a paragraph tag.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -36,7 +36,7 @@
       "devDependencies": {
         "@edx/frontend-build": "^11.0.2",
         "@edx/frontend-platform": "2.4.0",
-        "@edx/paragon": "^20.27.0",
+        "@edx/paragon": "^20.28.0",
         "@testing-library/dom": "^8.13.0",
         "@testing-library/react": "^12.1.1",
         "@testing-library/user-event": "^13.5.0",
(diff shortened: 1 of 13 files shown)


kenclary (approved):
I suspect this will lead to more css changes in the future. which is fine.

PR #226, opened 2023-02-01
fix: add check for question length for parser

KristinAoki:
This PR fixes the `endsWith` bug when a question was not defined or an empty string. Previously, the label parser assumed the question would have a length. Now, if the question is undefined or an empty string , it will not be parsed for labels.


kenclary (approved):
lgtm

PR #227, opened 2023-02-02
feat: fix feedback box bugs

KristinAoki:
JIRA Ticket: [TNL-10409](https://2u-internal.atlassian.net/browse/TNL-10409)

This PR fixes the following feedback bugs:

1. Previously when a user cleared the feedback fields, the section would auto-collapse. However the section was only expected to collapse when the toggle button is clicked. The behavior now matches expectations.
2. Previously when a user added feedback to a numeric or text input problem, the feedback would not save to the OLX. Now the feedback is properly saved and re-renders as expected. One thing to note is that numeric or text input problems only have one feedback field because their feedback tags do no support selected and unselected feedback.


kenclary (approved):
looks good

PR #228, opened 2023-02-02
fix: selectTypeModal header z-index

KristinAoki:
JIRA Ticket: [TNL-10407](https://2u-internal.atlassian.net/browse/TNL-10407)

This PR changes the z-index for the `SelectTypeModal` header. Previously, the header would be covered by the content in the main body when the page was scrolled. Now, the content scrolls below the header preventing, a visual conflict.


kenclary (approved):
yep

PR #231, opened 2023-02-03
feat: change title when problem type changes

rayzhou-bit:
This PR makes changes to problem type selection in the problem editor. When the problem type is changed through the problem type widget and the title of the block is a default value, the title will update to match the new problem type.

https://2u-internal.atlassian.net/browse/TNL-10399


kenclary (approved):
lgtm

PR #232, opened 2023-02-06
feat: update question parser to preserve text structure

KristinAoki:
JIRA Ticket: [TNL-10423](https://2u-internal.atlassian.net/browse/TNL-10423)

Previously, the xlm parser gave the question parser an object that grouped the html tags together and would strip out the ones that did not belong in the question. This caused the incoming xml to be rearranged when displayed in the question widget. Now the order of the text is preserved so that it renders as seen in studio.


kenclary (approved):
ok, I think this all makes sense.

PR #233, opened 2023-02-07
fix: settings col width in advanced view

KristinAoki:
JIRA Ticket: [TNL-10402](https://2u-internal.atlassian.net/browse/TNL-10402)

This PR fixes the width of the settings column in the advanced problem editor view. The raw editor was overriding the `flex-grow-1` class and requires its own width styling.

### DIFF
--- a/src/editors/containers/ProblemEditor/components/EditProblemView/__snapshots__/index.test.jsx.snap
+++ b/src/editors/containers/ProblemEditor/components/EditProblemView/__snapshots__/index.test.jsx.snap
@@ -8,8 +8,8 @@ exports[`EditorProblemView component renders raw editor 1`] = `
     className="editProblemView d-flex flex-row flex-nowrap justify-content-end"
   >
     <Container
-      className="mt-4.5"
-      size="lg"
+      className="advancedEditorTopMargin p-0"
+      fluid={true}
     >
       <RawEditor
         content={null}
--- a/src/editors/containers/ProblemEditor/components/EditProblemView/__snapshots__/index.test.jsx.snap
+++ b/src/editors/containers/ProblemEditor/components/EditProblemView/__snapshots__/index.test.jsx.snap
@@ -5,10 +5,11 @@ exports[`EditorProblemView component renders raw editor 1`] = `
   getContent={[Function]}
 >
   <div
-    className="editProblemView d-flex flex-row flex-nowrap justify-content-around"
+    className="editProblemView d-flex flex-row flex-nowrap justify-content-end"
   >
-    <div
+    <Container
       className="mt-4.5"
+      size="lg"
     >
       <RawEditor
         content={null}
@@ -18,13 +19,8 @@ exports[`EditorProblemView component renders raw editor 1`] = `
           }
         }
         lang="xml"
-        width={
-          Object {
-            "width": "75%",
-          }
-        }
       />
-    </div>
+    </Container>
     <span
       className="editProblemView-settingsColumn"
     >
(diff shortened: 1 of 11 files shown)


src/editors/containers/ProblemEditor/components/EditProblemView/index.jsx
```diff
@@ -24,17 +24,17 @@ export const EditProblemView = ({
...
-          {isAdvancedProblemType ? (
-            <RawEditor editorRef={editorRef} lang="xml" content={problemState.rawOLX} />
-          ) : (
-            <>
-              <QuestionWidget />
-              <AnswerWidget problemType={problemType} />
-            </>
-          )}
-        </span>
+      <div className="editProblemView d-flex flex-row flex-nowrap justify-content-around">
```
jesperhodge, line 27: I cannot say for sure, but I would say adding space around the content could cause problems (if the content is not big enough).

jesperhodge (changes requested):
On my end, the advanced editors look really weird. This may need some more fixing.
Screen Shot 2023-02-07 at 2 32 16 PM
Screen Shot 2023-02-07 at 2 31 46 PM

jesperhodge:
@KristinAoki The requirement from Jonathan is that the settings widget has a fixed width that doesn't change. I do not know why this doesn't work on the advanced editor, but we should keep this width. The question editor now needs to grow to fill the rest of the space.

jesperhodge (changes requested):
Hi, it looks much better. I still see three problems on advanced editor:
- The question editor is centered instead of stretched
- The settings widget is not as wide as it should be when the window width is reduced
- The top of the settings widget does not align with the top of the question editor

Screen Shot 2023-02-13 at 11 07 06 AM
Screen Shot 2023-02-13 at 11 06 40 AM

src/editors/containers/ProblemEditor/components/EditProblemView/__snapshots__/index.test.jsx.snap
```diff
@@ -7,8 +7,9 @@ exports[`EditorProblemView component renders raw editor 1`] = `
   <div
     className="editProblemView d-flex flex-row flex-nowrap justify-content-end"
   >
-    <span
-      className="flex-grow-1"
+    <Container
```
jesperhodge, line 10: I had a quick look why the layout was behaving so off. Basically the Container is the wrong element for this, it gives some behavior that we don't want at all. Firstly the container (or some other class responsible) has margin auto, which centers the advanced editor (which we don't want). Secondly it also has a max-width which is also not good, because we want to have it grow organically to fill the empty space. So flex-grow-1 together with no margin and no max-width fixes the problem of the container being centered weirdly.

src/editors/containers/ProblemEditor/components/EditProblemView/index.jsx
```diff
@@ -25,16 +27,16 @@ export const EditProblemView = ({
...
-          )}
-        </span>
+          </Container>
+        ) : (
+          <span className="flex-grow-1">
+            <QuestionWidget />
+            <AnswerWidget problemType={problemType} />
+          </span>
+        )}
         <span className="editProblemView-settingsColumn">
```
jesperhodge, line 40: I had a look at the actual reason the settings widget (or span container around it) gets too small. So it already has a (correct) flex-basis of 320px. Of course, that may shrink when the question editor expands in an unexpected way. So adding a `flex-shrink: 0` and maybe also `width: 320px` seems to keep it from shrinking so it stays correct. Additionally, there is some margin on the left side of the Settingswidget inside the flex element that takes up some of that 320px - we need to remove that and maybe add it in another place so we keep the correct margin from figma.
Where to make those changes: the class `editProblemView-settingsColumn` should be in the corresponding .scss file, which probably needs to be adjusted.

jesperhodge (changes requested):
I had a deeper look at what's going on here. There are some flex-issues that remain unaddressed which cause the faulty behavior. I left some comments on what is necessary to fix it, but it's probably easiest to have a call to get you up to speed or answer any questions I haven't answered.

jesperhodge (approved):

PR #234, opened 2023-02-07
feat: make radio and checkbox green

rayzhou-bit:
This PR makes the radio and checkboxes in problem editor green when checked. This is intended to override paragon.

https://2u-internal.atlassian.net/browse/TNL-10376

### DIFF
--- a/src/editors/containers/ProblemEditor/components/EditProblemView/index.scss
+++ b/src/editors/containers/ProblemEditor/components/EditProblemView/index.scss
@@ -3,3 +3,13 @@
     flex-basis: 320px;
   }
 }
+
+// overrides paragon in order to make checked radio and checkboxes green
+$checked-color: '%230D7D4D';
+
+.pgn__form-radio-input,
+.pgn__form-checkbox-input {
+  &:checked {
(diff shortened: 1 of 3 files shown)


src/editors/containers/ProblemEditor/components/EditProblemView/index.scss
```diff
@@ -2,4 +2,14 @@
   .editProblemView-settingsColumn {
     flex-basis: 320px;
   }
-}
\ No newline at end of file
+}
+
+// overrides paragon in order to make checked radio and checkboxes green
+$checked-color: '%230D7D4D';
```
jesperhodge, line 8: Can we open an issue or PR with paragon to fix this? This can be only a very temporary solution, because other themes have their own style and this will override them as well.

jesperhodge (changes requested):
The single select does not look like in figma for me locally. Maybe there is a way of getting it more in line with it? Figma screenshot:
Screen Shot 2023-02-07 at 3 18 42 PM

rayzhou-bit:
@jesperhodge Good catch on the radio buttons. I should have checked figma again.

jesperhodge (approved):

PR #235, opened 2023-02-07
fix: add empty answer to problem without answers

KristinAoki:
JIRA Ticket: [TNL-10428](https://2u-internal.atlassian.net/browse/TNL-10428)

Previously, when a user created a problem and saved it without adding an answer, they would be redirected to the advanced editor page because the parser did not check for an empty answer array. Now if a problem does not have answers, the parser adds an empty answer to the problem and takes the user to the visual editor.


jesperhodge (approved):
Amazing work on all of these PRs!

PR #240, opened 2023-02-10
fix: more correct parsing of settings, for scoring; use empty string instead of null for empty attempts. TNL-10433.

kenclary:
fix: more correct parsing of settings, for scoring; use empty string instead of null for empty attempts. TNL-10433.

### DIFF
--- a/src/editors/containers/ProblemEditor/data/SettingsParser.js
+++ b/src/editors/containers/ProblemEditor/data/SettingsParser.js
@@ -15,11 +15,15 @@ export const parseScoringSettings = (metadata) => {
   let scoring = {};
 
   let attempts = popuplateItem({}, 'max_attempts', 'number', metadata);
-  if (!_.isEmpty(attempts)) {
-    const unlimited = _.isNaN(attempts.number);
-    attempts = { ...attempts, unlimited };
-    scoring = { ...scoring, attempts };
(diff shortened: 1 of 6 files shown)


src/editors/containers/ProblemEditor/data/SettingsParser.js
```diff
@@ -15,11 +15,15 @@ export const parseScoringSettings = (metadata) => {
...
   let attempts = popuplateItem({}, 'max_attempts', 'number', metadata);
-  if (!_.isEmpty(attempts)) {
-    const unlimited = _.isNaN(attempts.number);
-    attempts = { ...attempts, unlimited };
-    scoring = { ...scoring, attempts };
+  if (_.isEmpty(attempts) || _.isNaN(attempts.number)) {
+    attempts = { unlimited: true, number: '' };
+  } else {
+    attempts.unlimited = false;
+    if (attempts.number < 0) {
```
KristinAoki, line 22: Should this same logic be applied to points? Right now I can add negative points.

kenclary: I have no idea; that's a question for the design.
This was just a case of finding another place in the code that should validate correct values for attempts, and we know that attempts shouldn't ever be negative.

KristinAoki (approved):

PR #242, opened 2023-02-10
fix: update problem description in preview block

KristinAoki:
The text that was rendered in the problem preview was the text for the answer description. Now the text matches the expected problem description found in the mocks.


connorhaugh (approved):
LGTM

PR #243, opened 2023-02-10
fix: getGeneralFeedback selectedFeedback error

KristinAoki:
This PR resolves the bug in the OLXParser inside the `getGeneralFeedback` function. The function expected that all answers would have `selectedFeedback` defined. However, when a single select or dropdown problem are opened without answers, `selectedFeedback` is not defined. This caused these problems to be rendered in the Advanced Editor when they should be rendered in the visual editor.


connorhaugh (approved):
:shipit: TYSM we love optional chaining!

PR #247, opened 2023-02-15
feat: remove Hints widget from advanced editor

rayzhou-bit:
This PR removes the Hints widget from the advanced editor. Hints shows up as a tag in the OLX so it should not be a widget.

https://2u-internal.atlassian.net/browse/TNL-10457


kenclary (approved):
nice and simple

PR #248, opened 2023-02-15
feat: move solution tag inside problem type tag

rayzhou-bit:
This PR moves the solution tag inside the problem type tag. When a problem block has multiple problems, each problem has their respective solutions. This change sets the user up for when they switch to the advanced editor.

https://2u-internal.atlassian.net/browse/TNL-10458


kenclary (approved):
looks good

PR #250, opened 2023-02-22
fix: check for if an answer changes to false

KristinAoki:
JIRA Ticket: [TNL-10445](https://2u-internal.atlassian.net/browse/TNL-10445)

This PR addresses a bug that prevented answers from clearing out for dropdown problems when multiple answers were selected a second time. Previously, the answers cleared out as expected when a user first changed from a multi-answer problem to a single answer problem. However, if the user reselected multiple answers and went back to a single answer problem, the answers were not deselected. Now, the user can go back and forth between the problem types with multiple answers selected and problem types with single answers with the answers clearing out when necessary.

### DIFF
--- a/src/editors/containers/ProblemEditor/components/EditProblemView/SettingsWidget/hooks.js
+++ b/src/editors/containers/ProblemEditor/components/EditProblemView/SettingsWidget/hooks.js
@@ -210,7 +210,9 @@ export const typeRowHooks = ({
     if (typeKey === ProblemTypeKeys.DROPDOWN) {
       if (correctAnswerCount > 1) {
         answers.forEach(answer => {
-          updateAnswer({ ...answer, correct: false });
+          if (answer.correct) {
+            updateAnswer({ ...answer, correct: false });
+          }
         });
       }
     }
--- a/src/editors/containers/ProblemEditor/components/EditProblemView/SettingsWidget/hooks.js
+++ b/src/editors/containers/ProblemEditor/components/EditProblemView/SettingsWidget/hooks.js
@@ -207,9 +207,13 @@ export const typeRowHooks = ({
   updateAnswer,
 }) => {
   const onClick = () => {
+    // Dropdown problems can only have one correct answer. When there is more than one correct answer
+    // from a previous problem type, the correct attribute for selected answers need to be set to false.
     if (typeKey === ProblemTypeKeys.DROPDOWN) {
+      // Check that previous problem had greater than one correct answer
       if (correctAnswerCount > 1) {
         answers.forEach(answer => {
+          // Check that the answer.correct is true and update answer.correct to false
           if (answer.correct) {
             updateAnswer({ ...answer, correct: false });
           }
--- a/src/editors/containers/ProblemEditor/components/EditProblemView/SettingsWidget/hooks.js
+++ b/src/editors/containers/ProblemEditor/components/EditProblemView/SettingsWidget/hooks.js
@@ -206,18 +206,19 @@ export const typeRowHooks = ({
   updateField,
   updateAnswer,
 }) => {
+  const clearPreviouslySelectedAnswers = () => {
+    answers.forEach(answer => {
+      if (answer.correct) {
+        updateAnswer({ ...answer, correct: false });
+      }
+    });
+  };
   const onClick = () => {
     // Dropdown problems can only have one correct answer. When there is more than one correct answer
     // from a previous problem type, the correct attribute for selected answers need to be set to false.
     if (typeKey === ProblemTypeKeys.DROPDOWN) {
-      // Check that previous problem had greater than one correct answer
       if (correctAnswerCount > 1) {
-        answers.forEach(answer => {
-          // Check that the answer.correct is true and update answer.correct to false
-          if (answer.correct) {
-            updateAnswer({ ...answer, correct: false });
-          }
-        });
+        clearPreviouslySelectedAnswers();
       }
     }
     if (blockTitle === ProblemTypes[problemType].title) {
--- a/src/editors/containers/ProblemEditor/components/EditProblemView/SettingsWidget/hooks.test.js
+++ b/src/editors/containers/ProblemEditor/components/EditProblemView/SettingsWidget/hooks.test.js
@@ -263,6 +263,10 @@ describe('Problem settings hooks', () => {
       {
         correct: true,
         id: 'b',
+      },
+      {
+        correct: false,
+        id: 'c',
       }];
       output = hooks.typeRowHooks({
         answers,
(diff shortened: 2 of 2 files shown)


src/editors/containers/ProblemEditor/components/EditProblemView/SettingsWidget/hooks.js
```diff
@@ -210,7 +210,9 @@ export const typeRowHooks = ({
     if (typeKey === ProblemTypeKeys.DROPDOWN) {
       if (correctAnswerCount > 1) {
         answers.forEach(answer => {
-          updateAnswer({ ...answer, correct: false });
+          if (answer.correct) {
+            updateAnswer({ ...answer, correct: false });
+          }
```
jesperhodge, line 215: I'm confused. I have seen that this fixes the issue, but I haven't the faintest idea why?

src/editors/containers/ProblemEditor/components/EditProblemView/SettingsWidget/hooks.test.js
```diff
@@ -263,6 +263,10 @@ describe('Problem settings hooks', () => {
       {
         correct: true,
         id: 'b',
+      },
+      {
+        correct: false,
+        id: 'c',
```
jesperhodge, line 269: If you add a test that fails when multiple answers stay selected in the dropdown but succeeds after your fix, this will prevent the same bug from reappearing due to future changes.

jesperhodge (changes requested):

src/editors/containers/ProblemEditor/components/EditProblemView/SettingsWidget/hooks.js
jesperhodge, line 215: Do you think extracting line 211-217 into a function named something like `clearPreviouslySelectedAnswers()` would be helpful to clarify what the code does? That would also help to reduce nested if statements for readability.

KristinAoki: I think it makes sense for me to add comments rather than a separate function based on the way that the code is currently structured for all the setting hooks.

src/editors/containers/ProblemEditor/components/EditProblemView/SettingsWidget/hooks.test.js
KristinAoki: The bug was in that the correctAnswerCount was negative so multiple answers stay selected because the count was never greater than one. I added a test that makes sure only answers with correct true are called because the underlying issue was updateAnswer being called extraneous times.

src/editors/containers/ProblemEditor/components/EditProblemView/SettingsWidget/hooks.test.js
```diff
@@ -278,6 +282,7 @@ describe('Problem settings hooks', () => {
       expect(setBlockTitle).toHaveBeenCalledWith('Dropdown');
       expect(updateAnswer).toHaveBeenNthCalledWith(1, { ...answers[0], correct: false });
       expect(updateAnswer).toHaveBeenNthCalledWith(2, { ...answers[1], correct: false });
+      expect(updateAnswer).not.toHaveBeenNthCalledWith(3, { ...answers[2], correct: false });
```
jesperhodge, line 285: good catch, this will catch the bug if it occurs again.

nit: if we find a bug, we want to regression test it. That would ideally mean having a separate test that explains what the bug is as part of the test name, and in general a test should only test one thing. Tests should also not rely on implementation details, since that leads to false positives and false negatives. However, fixing that is probably a bit too big of a scope for this PR.

src/editors/containers/ProblemEditor/components/EditProblemView/SettingsWidget/hooks.js
jesperhodge: nit: what about the structure of the code for the settings hooks prevents this? The settings hooks are all about extracting functions, and there is obviously a (very small) problem with the readability of the code here. If you don't like the extracted function in the module scope, a nice pattern in this case is just to declare the new function inside of the upper typeRowHooks function - we do that everywhere in these hooks files.
Code should be largely self-documenting and use explanatory variables, according to the clean code principles. In terms of clean code, comments are used to add extra information that does not restate what the code does, but clarifies intent or warns of consequences (for example that this bug may appear). Nested if-statements are usually not recommended because they increase the cognitive load for understanding what code does a lot (I usually avoid any level of nesting of if-statements.)

src/editors/containers/ProblemEditor/components/EditProblemView/SettingsWidget/hooks.js
```diff
@@ -207,10 +207,16 @@ export const typeRowHooks = ({
   updateAnswer,
 }) => {
   const onClick = () => {
+    // Dropdown problems can only have one correct answer. When there is more than one correct answer
+    // from a previous problem type, the correct attribute for selected answers need to be set to false.
     if (typeKey === ProblemTypeKeys.DROPDOWN) {
+      // Check that previous problem had greater than one correct answer
```
jesperhodge, line 213: This comment only restates what the line below says. If someone changes the code, they need to update the comment, and if the code is not clear than likely the comment is not clear. Better to only have the upper-level comment.

src/editors/containers/ProblemEditor/components/EditProblemView/SettingsWidget/hooks.js
```diff
@@ -207,10 +207,16 @@ export const typeRowHooks = ({
...
       if (correctAnswerCount > 1) {
         answers.forEach(answer => {
-          updateAnswer({ ...answer, correct: false });
+          // Check that the answer.correct is true and update answer.correct to false
```
jesperhodge, line 216: This comment only restates what the line below says. If someone changes the code, they need to update the comment, and if the code is not clear than likely the comment is not clear. Better to only have the upper-level comment.

src/editors/containers/ProblemEditor/components/EditProblemView/SettingsWidget/hooks.js
jesperhodge:
```suggestion
```

src/editors/containers/ProblemEditor/components/EditProblemView/SettingsWidget/hooks.js
jesperhodge:
```suggestion
```

jesperhodge (approved):
This looks good. I left 4 comments: 2 of them are nits, you can just read them and merge. Please don't feel obligated to change anything more unless you want to, it is probably out of scope for this PR.

PR #254, opened 2023-02-23
fix: file upload on safari

connorhaugh:
This PR switches out the form control select component for the dropdown component, as the result of this dropdown is to open a file upload dialog.

An upload dialog cannot be opened from a non-user action. 

A select's on change event is not a user action, so the component cannot be used.

Some minor styling changes may occur, but I tried to match the experiences as best I could, particularly re: a11y.

### DIFF
--- a/src/editors/containers/VideoEditor/components/VideoSettingsModal/components/TranscriptWidget/LanguageSelector.jsx
+++ b/src/editors/containers/VideoEditor/components/VideoSettingsModal/components/TranscriptWidget/LanguageSelector.jsx
@@ -113,36 +113,6 @@ export const LanguageSelector = ({
   );
 };
 
-/* <div className="col col-11 p-0">
-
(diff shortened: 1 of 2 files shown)


src/editors/containers/VideoEditor/components/VideoSettingsModal/components/TranscriptWidget/LanguageSelector.jsx
```diff
@@ -54,27 +59,57 @@ export const LanguageSelector = ({
...
   });
+  console.log({ localLang, language, openLanguages });
+
+  const getTitle = () => {
+    if (Object.prototype.hasOwnProperty.call(videoTranscriptLanguages, language)) {
+      return (
+        <ActionRow>
+          {videoTranscriptLanguages[language]}
+          <ActionRow.Spacer />
+          <Icon className="text-primary-500" src={Check} />
```
KristinAoki, line 70: Is there a reason for the check here? There isn't one in the mocks, only when you are looking at the dropdown list.

KristinAoki (approved):
Functionality is good on Safari and Chrome. Left one comment, a small ui request, but it is not blocking for merge.

src/editors/containers/VideoEditor/components/VideoSettingsModal/components/TranscriptWidget/LanguageSelector.jsx
connorhaugh: mmm I can give it a fix later if people get mad. Some paragon selects have them when you have an item selected.

PR #257, opened 2023-03-01
feat: expanding complex text area.

connorhaugh:
Prototype. This doesn't re-render the editor on blur, it simply hides it.


rayzhou-bit (changes requested):
I'm testing this in gallery view (as devstack is not working once again), so let me know if you're not seeing this on your end.

Issues:
- toolbar only seems to show when on a newline or when text is selected
- when adding images or other things like emojis/symbols/tables, they appear at the start of the textarea regardless of current cursor position
- selecting text and applying formatting does not work (code block does though)
- undo/redo doesn't work when you switch between different inline tinyMCEs. not sure if this is just a limitation of multiple tinyMCEs though.

QoL:
- it would be nice if toolbar appeared in a static place rather than following cursor similar to the demo shown here: https://www.tiny.cloud/docs/general-configuration-guide/use-tinymce-inline/

KristinAoki:
> * undo/redo doesn't work when you switch between different inline tinyMCEs. not sure if this is just a limitation of multiple tinyMCEs though.

this is expected because the redux is being updated `onBlur` so the editor doesn't recognize the previous changes when the editor was in focus.

KristinAoki:
> * toolbar only seems to show when on a newline or when text is selected
> QoL:
> 
> * it would be nice if toolbar appeared in a static place rather than following cursor similar to the demo shown here: https://www.tiny.cloud/docs/general-configuration-guide/use-tinymce-inline/

I talked with Jon about this and he preferred that the toolbar follows the cursor rather than having it in a fixed position outside of the editor.

rayzhou-bit (changes requested):
The formatting looks good now.
I'm still seeing an issue with adding images. Sometimes it'll add at cursor location, sometimes it won't.

Steps for when image does add at cursor location:
1. Type a couple lines of text.
2. Move cursor to a separate tinyMCE instance.
3. Move cursor back to the first tinyMCE instance.
4. Add the image.

When I directly try to add an image without doing the above steps, the image adds at the beginning of the textarea.

rayzhou-bit (approved):
This looks great!

PR #260, opened 2023-03-02
fix: parse explanation data correctly.

connorhaugh:
Explanation data was not being parsed because the test data was malformed-- the explanation actually goes INSIDE the input tag. This PR solves this issue in the parser, as well as it's test data.

https://2u-internal.atlassian.net/browse/TNL-10501
and https://2u-internal.atlassian.net/browse/TNL-10504


rayzhou-bit (approved):
Not sure how the previous code got past unit tests...
This looks good to me though.

PR #262, opened 2023-03-02
[BB-7157] Create new editor page for video upload

pkulkark:
## Description
This PR adds a new editor page for uploading videos in a course as per the [design doc](https://www.figma.com/file/cS42BX4OUsR8lOw6ppZHoz/Content-Libraries?node-id=3313%3A249596&t=XFyDHuxMZAZAiNZM-0).

## Supporting Information
This PR depends on the changes included in https://github.com/openedx/edx-platform/pull/31869 and https://github.com/openedx/frontend-app-course-authoring/pull/455.

OpenCraft Internal Jira task: [BB-7157](https://tasks.opencraft.com/browse/BB-7157)

## Screenshots
Screenshot from 2023-04-06 14-26-01

Screenshot from 2023-04-06 14-26-54

## Testing Information
1. Ensure your devstack contains the changes from 2 dependent PRs linked above.
2. See https://github.com/openedx/frontend-lib-content-components#how-to-set-up-development-workflow-of-v2-content-editors-in-studio-and-course-authoring-mfe for instructions on how to setup the video editor MFE.
3. Enable Video Uploads page by setting the feature flag `ENABLE_VIDEO_UPLOAD_PIPELINE` to true and setting VideoUploadsEnabledByDefault in the admin panel to enabled.
4. In Studio, select a course and choose `Video Uploads` from the dropdown options under `Content`.
5. You should see the new uploads page load as shown in the screenshot.
6. Drag and drop or click on Browse button to upload required files.
7. Check the console logs for any errors.


openedx-webhooks:
Thanks for the pull request, @pkulkark!  Please note that it may take us up to several weeks or months to complete a review and merge your PR.

Feel free to add as much of the following information to the ticket as you can:
- supporting documentation
- [Open edX discussion forum threads](https://discuss.openedx.org/)
- timeline information ("this must be merged by XX date", and why that is)
- partner information ("this is a course on edx.org")
- any other information that can help Product understand the context for the PR

All technical communication about the code itself will be done via the GitHub pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).


This is currently a draft pull request. When it is ready for our review and all tests are green, click "Ready for Review", or remove "WIP" from the title, as appropriate.

package.json
```diff
@@ -68,6 +68,7 @@
     "codemirror": "^6.0.0",
     "fast-xml-parser": "^4.0.10",
     "lodash-es": "^4.17.21",
+    "react-dropzone": "^14.2.3",
```
Cup0fCoffee, line 71: @pkulkark IIRC, you need to commit `package-lock.json` as well.

src/editors/containers/VideoUploadEditor/index.jsx
```diff
@@ -0,0 +1,117 @@
...
+  },
+  browse: {
+    defaultMessage: 'Browse files',
+    id: 'VideoUploadEditor.browse',
+  },
+  info: {
+    id: 'VideoUploadEditor.uploadInfo',
+    defaultMessage: 'Upload .mp4 or .mov files (5 GB max)',
+  }
+});
```
Cup0fCoffee, line 26: @pkulkark I think this can be extracted into `messages.js` by convention with the other editors.

src/editors/containers/VideoUploadEditor/index.jsx
```diff
@@ -0,0 +1,117 @@
...
+
+  const handleDrop = (files) => {
+    const extToMime = {
+      mp4: 'video/mp4',
+      mov: 'video/quicktime',
+    };
+    const supportedFormats = Object.keys(extToMime);
+
+    function getFileExtension(filename) {
+      return filename.slice((filename.lastIndexOf('.') - 1 >>> 0) + 2);
```
Cup0fCoffee, line 86: I'm not entierly sure about the purpose of `>>> 0` here (I could only imagine that it would be used to force a string to be converted to an int, but I don't think it's needed here). And then we do `- 1` and `+ 2`, which results in `+1` which removes the dot before the extension. I think this could be simplified to:
```suggestion
      return filename.slice(filename.lastIndexOf('.') + 1);
```
Or optionally be replaced as:
```suggestion
      return filerename.split('.').pop();
```

src/editors/containers/VideoUploadEditor/index.scss
```diff
@@ -0,0 +1,64 @@
+.dropzone-wrapper {
```
Cup0fCoffee, line 1: @pkulkark IIRC, the convention is to use `paragon` and `Bootstrap` classes as much as possible (within reason) and try to reduce the SCSS.

Cup0fCoffee (changes requested):
@pkulkark Looking good! Some minor changes. I've tested this locally, but it would be nice to be able to test it with the proper video upload pipeline, so looking forward to a sandbox instance.

Also, can you please run `npm run lint` and address linting issues related to your changes.

src/editors/containers/VideoUploadEditor/index.jsx
pkulkark: @Cup0fCoffee This was required because when you drag-and-drop multiple files, the dropzone returns `Blob` objects instead of `File` objects. And so the filenames get some additional characters and doesn't have the `type` parameter.

Cup0fCoffee: @pkulkark I don't think I understand. I've tried replicating on the devstack by drag and droping multiple files and once as well as doing them one by one, but I'm not seeing anything weird. And even it was, to me it seems like `filename.slice((filename.lastIndexOf('.') - 1 >>> 0) + 2)` and `filename.slice(filename.lastIndexOf('.') + 1)` do the same thing, because (assuming `lastIndexOf` doesn't throw and error because `filename` is `undefined` or `null`), `lastIndexOf` can only return an integer greater or equal to `-1`. So it's `n - 1`, `... >>> 0` doesn't do anything because it's a int already, and then `n - 1 + 2` -> `n + 1`.

package.json
pkulkark: @Cup0fCoffee You're right. I committed it now.

src/editors/containers/VideoUploadEditor/index.jsx
pkulkark: Done.

src/editors/containers/VideoUploadEditor/index.jsx
pkulkark: @Cup0fCoffee Sorry I got confused with another approach I had used before this. The purpose of `>>>0` was just to ensure that the value passed to slice() is never negative. I've updated it to use `Math.abs` to avoid any confusion (and also fix the lint issue about using bitwise operators).

pkulkark:
As discussed in the jira ticket ([BB-7157](https://tasks.opencraft.com/browse/BB-7157?focusedCommentId=266176&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-266176)), putting the findings here:

I tried replacing my component with [edx/paragon](https://github.com/openedx/paragon)'s Dropzone component but it turns out edx/paragon's implementation is quite buggy. It doesn't allow multiple file uploads, doesn't apply the className property correctly so it's not possible to style it and the file selector (when you click on the component) doesn't work. Furthermore, it is actually implementing it in pretty much the same way that my component is (i.e. using react-dropzone). So we have two ways to go with this:
1. Drop edx/paragon's component and continue with the initial implementation (what you see in the PR). The only thing would be that we won't have proper snapshot tests for it (I suspect it would be the same for paragon but I haven't tested it yet).
2. Focus on fixing the issues with edx/paragon's implementation and then use that.

src/editors/containers/VideoUploadEditor/hooks.js
```diff
@@ -0,0 +1,42 @@
...
+    data.files.push({
+      file_name: file.name,
+      content_type: file.type,
+    });
+  });
+  dispatch(requests.uploadVideo({
+    data,
+    onSuccess: (response) => {
+      const { files } = response.json();
+      Object.values(files).forEach((fileObj) => {
```
giovannicimolin, line 15: @pkulkark This will be used when a user wants to add a single video to a course, so it shouldn't allow multiple file uploads.

For context, the user flow will be:
1. From studio, a user wants to add a video to a course.
2. User directly clicks on the video block in a unit, and is taken to a gallery page to browse available videos (being worked on a separate task).
3. Uses decides to upload a new video from that gallery, click the upload button and is taken to the page you are implementing here.
4. He either uploads a **single video** or adds a youtube URL.
5. Once the upload is complete, the user is taken to the video edit page with the information pre-filled there.

For your task, the only relevant item is (4) - building the upload page.

giovannicimolin:
@pkulkark After [this discussion](https://openedx.slack.com/archives/C04PPHQK1UM/p1680623433934969) on Slack, we decided to use the approach 1: keep using our component here to unblock follow-up tasks.

At the end of the project, we'll add an extra "upstreaming to paragon" task to move to paragon's component and fix its bugs.

pkulkark:
Thanks for the update and context @giovannicimolin. That makes sense.

> It doesn't allow multiple file uploads

I meant to say this was one of the issues with the edx/paragon's implementation of the Dropzone component, but not relevant to our use case here. The underlying react-dropzone allows it with a flag. So ideally edx/paragon's component should also have the option.

src/editors/containers/VideoUploadEditor/hooks.js
pkulkark: Ack. Will remove the flag and update the logic.

Cup0fCoffee: @pkulkark Do you think this function should be simplified to handle only one file upload?

src/editors/containers/VideoUploadEditor/hooks.js
```diff
@@ -0,0 +1,42 @@
...
+        const fileName = fileObj.file_name;
+        const uploadUrl = fileObj.upload_url;
```
Cup0fCoffee, line 17: FYI, you can do this, but I don't know which one is better:
```suggestion
        const { file_name: fileName, upload_url: uploadUrl } = fileObj;
```

src/editors/containers/VideoUploadEditor/hooks.test.js
```diff
@@ -0,0 +1,68 @@
...
+  it('should dispatch uploadVideo action with correct data and onSuccess callback', () => {
+    const data = {
+      files: [
+        { file_name: 'file1.mp4', content_type: 'video/mp4' },
+        { file_name: 'file2.mov', content_type: 'video/quicktime' },
+      ],
+    };
+    const onSuccess = jest.fn();
+    hooks.uploadVideo({ dispatch, supportedFiles });
+    expect(dispatch).toHaveBeenCalledWith(requests.uploadVideo({ data, onSuccess }));
```
Cup0fCoffee, line 26: I have doubts this works as intended.

src/editors/supportedEditors.js
```diff
@@ -1,6 +1,7 @@
 import TextEditor from './containers/TextEditor';
 import VideoEditor from './containers/VideoEditor';
 import ProblemEditor from './containers/ProblemEditor';
+import VideoUploadEditor from './containers/VideoUploadEditor';
```
Cup0fCoffee, line 4: nit: I don't know what qualifies as "ADDED_EDITORS", but since in the `supportedEditors` object the new editor was added below the comment, to be consistent, the corresponding import should be below the comment on line 6 as well.

Cup0fCoffee (changes requested):
@pkulkark A few comments from the initial review. I have some suggestions in regards to tests which I will post separately.

src/editors/containers/VideoUploadEditor/hooks.js
pkulkark: @Cup0fCoffee The response here is from the edx-platform API, which returns a list of files, as the API supports multiple file uploads. See https://github.com/openedx/edx-platform/blob/master/cms/djangoapps/contentstore/views/videos.py#L724-L742. Although in this case, it would return a list containing a single file, I left it as is in the interest of time (also didn't think it would affect computationally - let me know if I'm wrong about that).

Cup0fCoffee: I haven't realized that. Thank you for clarifying. In that case I think it's ok.

src/editors/containers/VideoUploadEditor/hooks.test.js
Cup0fCoffee: @pkulkark As far as I know, `requests.uploadVideo({ data, onSuccess })` is going to return `undefined`, since it's mocked, so it doesn't really test anything. I noticed that this pattern is used all over this code base, and imho it's a bad pattern and should be avoided.

src/editors/containers/VideoUploadEditor/index.scss
pkulkark: @Cup0fCoffee Done. I moved most of it out of the SCSS file. Only kept a few custom styling.

src/editors/containers/VideoUploadEditor/hooks.test.js
pkulkark: @Cup0fCoffee Wouldn't the test fail in that case? I wanted to mock the studio API call. I admin my js testing skills is not very good. How would you suggest this be changed?

src/editors/supportedEditors.js
pkulkark: Ack. Moved it.

src/editors/containers/VideoUploadEditor/hooks.test.js
Cup0fCoffee: @pkulkark You can test it by replacing the `requests.uploadVideo({ data, onSuccess })` with `undefined` and running the tests - they will pass, because the mock doesn't have an implementation.

src/editors/containers/VideoUploadEditor/index.test.jsx
```diff
@@ -0,0 +1,113 @@
...
+  it('renders without errors', () => {
+    const { container } = renderEditorComponent();
+    expect(container).toMatchSnapshot();
+  });
+
+  it('renders the input field for pasting the video ID or URL', () => {
+    const { getByPlaceholderText } = renderEditorComponent();
+    const input = getByPlaceholderText('Paste your video ID or URL');
+    expect(input).toBeInTheDocument();
+  });
```
Cup0fCoffee, line 50: @pkulkark I think this test is redundant.

src/editors/containers/VideoUploadEditor/index.test.jsx
```diff
@@ -0,0 +1,113 @@
...
+  it('shows error message with unsupported files', async () => {
+    const { getByTestId, findByText } = renderEditorComponent();
+    const fileInput = getByTestId('fileInput');
+
+    const unsupportedFile = new File(['(⌐□_□)'], 'unsupported.avi', { type: 'video/avi' });
+    fireEvent.change(fileInput, { target: { files: [unsupportedFile] } });
+
+    const errorMsg = await findByText('Video must be an MP4 or MOV file');
+    expect(errorMsg).toBeInTheDocument();
+  });
```
Cup0fCoffee, line 69: @pkulkark Can you add a similar test, but with a supported file?

src/editors/containers/VideoUploadEditor/index.test.jsx
```diff
@@ -0,0 +1,113 @@
...
+
+  it('handles unsupported video file formats', () => {
+    const { getByTestId } = renderUploaderComponent();
+    const fileInput = getByTestId('fileInput');
+    const unsupportedFile = new File(['(⌐□_□)'], 'unsupported.avi', { type: 'video/avi' });
+
+    fireEvent.change(fileInput, { target: { files: [unsupportedFile] } });
+    expect(mockOnUpload).not.toHaveBeenCalled();
+  });
```
Cup0fCoffee, line 95: @pkulkark Does that actually call the `onDrop` call back? I don't see the logic in the `VideoUploader` which would filter the files by type. I think it passes them to the `onUpload` callback, which does the filtering. The fact that this test is passing means that `.change` doesn't trigger the `onDrop`.

src/editors/containers/VideoUploadEditor/index.test.jsx
```diff
@@ -0,0 +1,113 @@
...
+      fireEvent.drop(dropzone, {
+        dataTransfer: {
+          files: [file],
+          types: ['Files'],
+        },
+      });
+    });
+
+    expect(mockOnUpload).toHaveBeenCalledWith(file);
+  });
```
Cup0fCoffee, line 112: @pkulkark  Can we also check if the selected file displayed in the UI?

src/editors/containers/VideoUploadEditor/index.jsx
```diff
@@ -0,0 +1,129 @@
...
+  const [textInputValue, setTextInputValue] = useState('');
+
+  const { getRootProps, getInputProps, isDragActive } = useDropzone({
+    accept: 'video/*',
+    multiple: false,
+    onDrop: (acceptedFiles) => {
+      if (acceptedFiles.length > 0) {
+        const uploadfile = acceptedFiles[0];
+        setUploadedFile(uploadfile);
+        onUpload(uploadfile);
```
Cup0fCoffee, line 23: @pkulkark Can this create a scenario when the file is set, but hasn't been uploaded? I guess it's ok if the parent component displays the error, and the user can remove the selected file and choose a different one. Just pointing it out, in case it's not an intended behavior.

Cup0fCoffee (approved):
:+1: Have a couple of comments/question in regards to the tests and some logic, but ok otherwise.

- [x] I tested this
- [x] I read through the code
- [ ] I checked for accessibility issues
- [ ] Includes documentation
- [ ] I made sure any change in configuration variables is reflected in the corresponding client's `configuration-secure` repository.

src/editors/containers/VideoUploadEditor/index.test.jsx
pkulkark: Ack. Will remove it.

src/editors/containers/VideoUploadEditor/index.test.jsx
pkulkark: Sure, will add it.

src/editors/containers/VideoUploadEditor/index.test.jsx
pkulkark: You're right. It doesn't. It can't really find the element because react-dropzone hides it. I'll remove it since we are covering it in `VideoUploadEditor` above.

src/editors/containers/VideoUploadEditor/index.test.jsx
pkulkark: I removed displaying part because it was not in the designs. I think they'll be handled in a separate work flow.

src/editors/containers/VideoUploadEditor/index.jsx
pkulkark: This will be needed when we implement the displaying part. I think it requires a thumbnail of the video to be shown which should be handled in a later task.

src/editors/containers/VideoUploadEditor/index.test.jsx
pkulkark: Done.

src/editors/containers/VideoUploadEditor/index.test.jsx
pkulkark: Done.

src/editors/containers/VideoUploadEditor/index.test.jsx
pkulkark: Done.

kenclary (approved):
looks good

openedx-webhooks:
@pkulkark 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/frontend-lib-content-components&entry.867055334=https://github.com/openedx/frontend-lib-content-components/pull/262&entry.1484655318=https://github.com/pkulkark&entry.752974735=2023-03-02+20:50&entry.1917517419=2023-04-26+17:41&entry.2133058324=Yes) so we can improve your experience in the future.

PR #268, opened 2023-03-08
fix: ProblemEditor useEffect at beginning of component; message for block failed to load. TNL-10507

kenclary:
fix: ProblemEditor useEffect at beginning of component; message for block failed to load. TNL-10507

### DIFF
--- a/src/editors/containers/ProblemEditor/index.jsx
+++ b/src/editors/containers/ProblemEditor/index.jsx
@@ -2,22 +2,30 @@ import React from 'react';
 import { connect } from 'react-redux';
 import PropTypes from 'prop-types';
 import { Spinner } from '@edx/paragon';
-import { injectIntl } from '@edx/frontend-platform/i18n';
+import { injectIntl, FormattedMessage } from '@edx/frontend-platform/i18n';
 import SelectTypeModal from './components/SelectTypeModal';
 import EditProblemView from './components/EditProblemView';
 import { selectors, thunkActions } from '../../data/redux';
 import { RequestKeys } from '../../data/constants/requests';
+import messages from './messages';
 
 export const ProblemEditor = ({
   onClose,
   // Redux
   problemType,
   blockFinished,
+  blockFailed,
   studioViewFinished,
   blockValue,
   initializeProblemEditor,
   assetsFinished,
 }) => {
+  React.useEffect(() => {
+    if (blockFinished && studioViewFinished && assetsFinished && !blockFailed) {
+      initializeProblemEditor(blockValue);
+    }
+  }, [blockFinished, studioViewFinished, assetsFinished, blockFailed]);
+
   if (!blockFinished || !studioViewFinished || !assetsFinished) {
     return (
       <div className="text-center p-6">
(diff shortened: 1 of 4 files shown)


src/editors/containers/ProblemEditor/index.jsx
```diff
@@ -2,22 +2,26 @@ import React from 'react';
...
   // Redux
   problemType,
   blockFinished,
+  blockFailed,
   studioViewFinished,
   blockValue,
   initializeProblemEditor,
   assetsFinished,
 }) => {
+  React.useEffect(() => initializeProblemEditor(blockValue), []);
```
jesperhodge, line 23: Now that I am paying attention to the specifics: I think the reason that the useEffect was put after the return statement is to not call the `initializeProblemEditor` method unless block, studioview, and assets finished. The comment says "once data is loaded, init store". It was just the wrong approach. How to call the `initializeProblemEditor` method  conditionally is to add the condition to the callback method inside of the useEffect. However, for the hook to register that the booleans have changed, you also need to put them into the dependency array.

jesperhodge (changes requested):
See my comment; also, there are tests for the new functionality missing which is why the pipeline fails.

src/editors/containers/ProblemEditor/index.jsx
```diff
@@ -29,9 +37,14 @@ export const ProblemEditor = ({
       </div>
     );
   }
-  // once data is loaded, init store
-  React.useEffect(() => initializeProblemEditor(blockValue), []);
-  // TODO: INTL MSG, Add LOAD FAILED ERROR using BLOCKFAILED
+
+  if (blockFailed) {
+    return (
+      <div className="text-center p-6">
```
jesperhodge, line 43: Should we use a Toast like in the TextEditor, just for consistency?

src/editors/containers/ProblemEditor/index.jsx
```diff
@@ -2,22 +2,30 @@ import React from 'react';
...
   // Redux
   problemType,
   blockFinished,
+  blockFailed,
   studioViewFinished,
   blockValue,
   initializeProblemEditor,
   assetsFinished,
 }) => {
+  React.useEffect(() => {
```
jesperhodge, line 23: `initializeProblemEditor` seems to be a line that is not covered by tests - probably this is what codecov is complaining about? Up to now we usually have useEffects and such in the hooks file and then tested by mocking out effects and such. That could be a solution here if you don't want to use new tooling.

I'm just trying to suggest anything that could fix the code coverage problem, but go ahead and fix that in any way you see fit, I don't have any preference.
- Wile the place where this useEffect logic is right now is not consistent with the rest of the application, that is not due to you and not something we need to fix now.

kenclary: isn't "test `initializeProblemEditor`" the whole point of the (currently excluded) first bullet in the ticket?

jesperhodge (approved):
The comments I left earlier are just for the pipeline and such, so this is approved from my side, just needs the pipeline figured out and so on.

PR #272, opened 2023-03-10
feat: correct answer feedback widget

rayzhou-bit:
OLX for numeric input problem type only accepts one feedback for correct answers. This PR makes changes to the problem editor to follow this behavior by making these changes for the Numeric Input problem type:

- Creating a Correct Answer Feedback widget in the widget settings
- Ensuring OLX → react state parses correctly
- Ensuring react state → OLX parses correctly
- Removing inline selected feedback and general feedback widget

Feedback functionality for Numeric Input problem type is detailed in RTDocs: https://edx.readthedocs.io/projects/edx-partner-course-staff/en/latest/exercises_tools/numerical_input.html#adding-feedback-to-a-numerical-input-problem

https://2u-internal.atlassian.net/browse/TNL-10524

### DIFF
--- a/src/editors/containers/ProblemEditor/data/OLXParser.js
+++ b/src/editors/containers/ProblemEditor/data/OLXParser.js
@@ -226,7 +226,6 @@ export class OLXParser {
 
   parseNumericResponse() {
     const { numericalresponse } = this.problem;
-    let answerFeedback = '';
     const answers = [];
     let responseParam = {};
     // TODO: UI needs to be added to support adding tolerence in numeric response.
@@ -242,7 +241,6 @@ export class OLXParser {
       id: indexToLetterMap[answers.length],
       title: numericalresponse['@_answer'],
       correct: true,
-      selectedFeedback: feedback,
       ...responseParam,
     });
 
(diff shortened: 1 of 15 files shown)


src/editors/containers/ProblemEditor/data/OLXParser.js
```diff
@@ -242,32 +241,27 @@ export class OLXParser {
...
-      answerFeedback = this.getFeedback(additionalAnswer);
       answers.push({
         id: indexToLetterMap[answers.length],
         title: additionalAnswer['@_answer'],
         correct: true,
-        selectedFeedback: answerFeedback,
       });
     }
-    return { answers };
+    return { answers, correctAnswerFeedback: feedback };
```
KristinAoki, line 264: Is the feedback supposed to be applied to all the correct answers?

KristinAoki (commented):
The functionality looks good to me. When I added the correct feedback it was only applied to the first correct answer, not all of them.

KristinAoki:
I read on your [feedback matrix](https://docs.google.com/spreadsheets/d/1NM3dYct3EBU40Juk4mrw5lRO9_sXpzLzy491OPw_PeA/edit#gid=0) and noticed that it said that numeric inputs can only have one feedback is allowed regardless of the number of answers. However when I looked at the linked RTD it says:

> If you define multiple correct responses, you can define feedback for each response.

Wouldn't this mean that inline feedback should be allowed?

rayzhou-bit:
@KristinAoki Thank you for catching this. When I was testing around with the OLX on stage, I thought I only saw one feedback for numeric. However, I just tried again and it does accept different feedback for different answers.

PR #273, opened 2023-03-13
feat: remove general and group feedback

rayzhou-bit:
To match how the OLX currently works, this PR removes general feedback and group feedback for problem types except for multi-select.

https://2u-internal.atlassian.net/browse/TNL-10523


KristinAoki (approved):
If we are not using the `GeneralFeedback` widget, do we want to delete the files? Or are we keeping them for possible future work?

PR #274, opened 2023-03-13
feat: update text areas to expandable text areas

KristinAoki:
JIRA Ticket: [TNL-10496](https://2u-internal.atlassian.net/browse/TNL-10496)

This PR changes the following text areas to expandable text areas:

- [x] Answers
- [x] Solution
- [x] Feedback
- [x] Hints
- [x] Group Feedback

Each of the changes is in their own commit.

The expandable text areas will allow the users to add rich text to their problems to match the freedom of using markdown in the legacy experience.

### DIFF
--- a/src/editors/containers/ProblemEditor/components/EditProblemView/AnswerWidget/__snapshots__/AnswerOption.test.jsx.snap
+++ b/src/editors/containers/ProblemEditor/components/EditProblemView/AnswerWidget/__snapshots__/AnswerOption.test.jsx.snap
@@ -101,12 +101,13 @@ exports[`AnswerOption render snapshot: renders correct option with numeric input
   <div
     className="ml-1 flex-grow-1"
   >
-    <ExpandableTextArea
-      error={false}
-      errorMessage={null}
-      id="answer-A"
+    <Form.Control
+      as="textarea"
+      autoResize={true}
+      className="answer-option-textarea text-gray-500 small"
+      onChange={[Function]}
       placeholder="Enter an answer"
-      setContent={[Function]}
+      rows={1}
       value="Answer 1"
     />
     <Body>
--- a/src/editors/containers/ProblemEditor/data/ReactStateOLXParser.js
+++ b/src/editors/containers/ProblemEditor/data/ReactStateOLXParser.js
@@ -96,8 +96,8 @@ class ReactStateOLXParser {
       const feedback = [];
       let singleAnswer = {};
       const title = answerTitles ? this.parser.parse(answerTitles[answer.id]) : { '#text': answer.title };
-      const currentSelectedFeedback = selectedFeedback ? selectedFeedback[answer.id] : null;
-      const currentUnselectedFeedback = unselectedFeedback ? unselectedFeedback[answer.id] : null;
+      const currentSelectedFeedback = selectedFeedback?.[answer.id] || null;
+      const currentUnselectedFeedback = unselectedFeedback?.[answer.id] || null;
       let isEmpty;
       if (answerTitles) {
         isEmpty = Object.keys(title)?.length <= 0;
@@ -150,9 +150,12 @@ class ReactStateOLXParser {
   addGroupFeedbackList() {
     const compoundhint = [];
     const { groupFeedbackList } = this.problemState;
+    const { groupFeedback } = this.editorObject;
     groupFeedbackList.forEach((element) => {
+      const feedbackString = groupFeedback?.[element.id];
+      const parsedFeedback = this.parser.parse(feedbackString);
       compoundhint.push({
-        '#text': element.feedback,
+        ...parsedFeedback,
         '@_value': element.answers.join(' '),
       });
     });
@@ -237,7 +240,7 @@ class ReactStateOLXParser {
     const wrongAnswers = [];
     let firstCorrectAnswerParsed = false;
     answers.forEach((answer) => {
-      const correcthint = this.getAnswerHints(selectedFeedback[answer.id]);
+      const correcthint = this.getAnswerHints(selectedFeedback?.[answer.id]);
       if (this.hasAttributeWithValue(answer, 'title')) {
         if (answer.correct && firstCorrectAnswerParsed) {
           additionAnswers.push({
@@ -303,7 +306,7 @@ class ReactStateOLXParser {
     const additionalAnswers = [];
     let firstCorrectAnswerParsed = false;
     answers.forEach((answer) => {
-      const correcthint = this.getAnswerHints(selectedFeedback[answer.id]);
+      const correcthint = this.getAnswerHints(selectedFeedback?.[answer.id]);
       if (this.hasAttributeWithValue(answer, 'title')) {
         if (answer.correct && !firstCorrectAnswerParsed) {
           firstCorrectAnswerParsed = true;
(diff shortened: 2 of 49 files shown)


src/editors/containers/ProblemEditor/data/ReactStateOLXParser.js
```diff
@@ -4,6 +4,10 @@ import { ProblemTypeKeys } from '../../../data/constants/problem';
 
 class ReactStateOLXParser {
   constructor(problemState) {
+    const parserOptions = {
+      ignoreAttributes: false,
+      alwaysCreateTextNode: true,
```
connorhaugh, line 9: What would require the parser & builder to have different sets of options here? builderOptions and parserOptions are slightly different.

src/editors/containers/ProblemEditor/data/ReactStateOLXParser.js
```diff
@@ -134,7 +139,8 @@ class ReactStateOLXParser {
   }
 
   addQuestion() {
-    const { question } = this.editorObject || this.problemState;
```
connorhaugh, line 137: why remove this object dereferencing?  your change can still produce null as a result?

src/editors/containers/ProblemEditor/data/mockData/editorTestData.js
```diff
@@ -0,0 +1,62 @@
...
+  question: '<p>You can use this template as a guide to the simple editor markdown and OLX markup to use for checkboxes with hints and feedback problems. Edit this component to replace this template with your own assessment.</p><label>Add the question text, or prompt, here. This text is required.</label><em>You can add an optional tip or note related to the prompt like this.</em>',
+  answers: {
+    A: 'a correct answer',
+    B: 'an incorrect answer',
+    C: 'an incorrect answer',
+    D: 'a correct answer',
+  },
+};
+
+// export const multipleChoiceWithoutAnswers = {
```
connorhaugh: what purpose does this comment serve? I think you may want to remove it?

src/editors/containers/ProblemEditor/data/mockData/editorTestData.js
```diff
@@ -0,0 +1,62 @@
...
+//   solution: '',
+//   question: '',
+//   answers: { A: '' },
+// };
+
+export const dropdownWithFeedbackAndHints = {
+  solution: '',
```
connorhaugh, line 41: why do only come of these have solutions?

src/editors/sharedComponents/TinyMceWidget/hooks.js
```diff
@@ -57,12 +57,18 @@ export const replaceStaticwithAsset = ({
   imageSrcs.forEach(src => {
     const currentContent = content;
     let staticFullUrl;
-    if (src.startsWith('/static/') && imageUrls.length > 0) {
-      const imgName = src.substring(8, src.indexOf('"'));
+    const isStatic = src.startsWith('/static/');
+    const isExpandableAsset = src.startsWith('/assets/') && editorType === 'expandable';
```
connorhaugh, line 61: Can you explain the logic of what you are doing here in words? I am really bad at reading regex

src/editors/containers/ProblemEditor/data/mockData/editorTestData.js
KristinAoki: I based on the data that was already present in `olxTestData`

src/editors/sharedComponents/TinyMceWidget/hooks.js
KristinAoki: I am removing everything except the image name. I make sure that the replacement starts at the begin file path and matches all characters until the last slash/

src/editors/containers/ProblemEditor/data/ReactStateOLXParser.js
KristinAoki: In a later commit, I change it. I also just realized that question should always be defined so I can change it to
```
const { question } = this.editorObject
```

src/editors/containers/ProblemEditor/data/ReactStateOLXParser.js
KristinAoki: I am not sure. I based this on how options are separated in `OLXParser`

connorhaugh: Ok i think I understand.

src/editors/containers/ProblemEditor/components/EditProblemView/AnswerWidget/__snapshots__/AnswerOption.test.jsx.snap
connorhaugh, line 1: we should have a test in this file where the problem is multiselect or singleselect.

KristinAoki: There is one, but it is not new in PR so it will not show in the diffs

rayzhou-bit (changes requested):
Overall, the code looks really good!

I noticed two things:
- I missed this in the component review. After you add an image, when you click on it, it offers you 3 options for justifying. Clicking left or right justify bugs out the expandable textarea. I imagine these options shouldn't exist as there is no option to justify the text.
- Explanation in the Show Answer widget does not persist when the widget is collapsed and re-opened. Not allowing collapse would fix this is my guess.

An additional thought I had that might be a question for Jon: If we don't allow collapse for certain widgets, should these widgets look different (since the open/collapse arrow looks clickable but is not)?

connorhaugh (changes requested):
Just a miss-merge I think.

src/editors/data/redux/thunkActions/problem.js
```diff
@@ -3,14 +3,10 @@ import { actions } from '..';
 import { OLXParser } from '../../../containers/ProblemEditor/data/OLXParser';
 import { parseSettings } from '../../../containers/ProblemEditor/data/SettingsParser';
 import { ProblemTypeKeys } from '../../constants/problem';
-import ReactStateOLXParser from '../../../containers/ProblemEditor/data/ReactStateOLXParser';
 import { blankProblemOLX } from '../../../containers/ProblemEditor/data/mockData/olxTestData';
 
-export const switchToAdvancedEditor = () => (dispatch, getState) => {
```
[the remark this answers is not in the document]
KristinAoki: This change is related to my changes because when you switch from the visual to advanced editor it is parsing the current content. Earlier we determined that the rawOLX was not updating so I thought it was better to remove the code then add a new line that doesn't really contribute anything.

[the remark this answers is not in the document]
KristinAoki: However, after looking back at the code again, I figured out why the rawOLX was not updating when switching from the visual editor to the advanced editor. The variable that we were calling in `updateField` was `rawOlx`; however, the variable's name in the redux store is `rawOLX`. 🤦‍♀️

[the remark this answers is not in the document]
connorhaugh: ugh that's brutal glad you caught that!

rayzhou-bit (approved):
looks good to me!

PR #276, opened 2023-03-14
fix: show answer attempts spacing

connorhaugh:
added a margin to meet requirements. it is 4 not three to handle the floating label on the input.


kenclary (approved):
good

Issue #279, opened 2023-03-16
Upgrade Node JS from 16 to 18


abdullahwaheed:
**Abstract**
Node 16's support will end on [September 11, 2023](https://nodejs.org/en/blog/announcements/nodejs16-eol/). So we have to upgrade node JS to v18 before that.

**Extras**
Project should use `.nvmrc` to have a consistent node version
We have to update Readme accordingly.

[Upgrade Node JS from 16 to 18](https://github.com/openedx/platform-roadmap/issues/243)

jesperhodge:
Great issue. Let's consider doing this at the same time we are upgrading node to v18 on projects that are using this library, for example, frontend-app-course-authoring. Just so we don't create any emergent problems by possibly having different node versions.

abdullahwaheed:
yes we are planning to upgrade all the MFEs at once. Our team is creating draft PRs in all the repos and can be merged at the same time. e.g: #281 and https://github.com/openedx/frontend-app-course-authoring/pull/464

PR #281, opened 2023-03-16
feat: upgraded to node v18, added .nvmrc and updated workflows

BilalQamar95:
## Ticket
[Upgrade Node Js from 16 to 18](https://github.com/openedx/frontend-lib-content-components/issues/279)

## Description
- Added support for node v18, updated .nvmrc and workflows. 
- Updated `frontend-platform` & `frontend-build` versions


jesperhodge:
This is great!
I have two notes on it we should think about.
1) there's another PR for course-authoring to port that to v18. I think both should be tested and merged together, so we don't have any version mismatch.
2) we are beta testing the new problem editor right now. I'm not very worried about breaking something, but still think it may be better practice to work iteratively and not introduce new changes unnecessarily before launch. How do you feel about doing the version upgrade after the problem editor launch? That way we can test it a bit more separately.

KristinAoki (approved):

PR #285, opened 2023-03-20
feat: add default advanced setting ui callouts

KristinAoki:
JIRA Ticket: [TNL-10429](https://2u-internal.atlassian.net/browse/TNL-10429)

This PR adds "(Default)" to the attempt field and show answer field when the user selects/enters the default value defined in the courses advanced settings. Additionally this PR sets the attempt field, show answer field, and show reset button value to the course's default value when creating a new problem.

### DIFF
--- a/src/editors/containers/ProblemEditor/components/EditProblemView/SettingsWidget/hooks.js
+++ b/src/editors/containers/ProblemEditor/components/EditProblemView/SettingsWidget/hooks.js
@@ -136,7 +136,7 @@ export const scoringCardHooks = (scoring, updateSettings, defaultValue) => {
     let attemptNumber = parseInt(event.target.value);
     const { value } = event.target;
     if (_.isNaN(attemptNumber)) {
-     if (value === '') {
+      if (value === '') {
         attemptNumber = defaultValue;
         setAttemptDisplayValue(`${defaultValue} (Default)`);
       } else {
@@ -144,7 +144,7 @@ export const scoringCardHooks = (scoring, updateSettings, defaultValue) => {
         unlimitedAttempts = true;
       }
     } else if (attemptNumber <= 0) {
-        attemptNumber = 0;
+      attemptNumber = 0;
     } else if (attemptNumber === defaultValue) {
       const attemptNumberStr = value.replace(' (Default)');
       attemptNumber = parseInt(attemptNumberStr);
--- a/src/editors/containers/ProblemEditor/components/EditProblemView/SettingsWidget/hooks.js
+++ b/src/editors/containers/ProblemEditor/components/EditProblemView/SettingsWidget/hooks.js
@@ -10,6 +10,7 @@ export const state = {
   cardCollapsed: (val) => useState(val),
   summary: (val) => useState(val),
   showAttempts: (val) => useState(val),
+  local: (val) => useState(val),
 };
 
 export const showAdvancedSettingsCards = () => {
@@ -120,17 +121,33 @@ export const resetCardHooks = (updateSettings) => {
   };
 };
 
-export const scoringCardHooks = (scoring, updateSettings) => {
+export const scoringCardHooks = (scoring, updateSettings, defaultValue) => {
+  const [local, setLocal] = module.state.local(scoring.attempts.number);
   const handleMaxAttemptChange = (event) => {
     let unlimitedAttempts = false;
     let attemptNumber = parseInt(event.target.value);
+    const { value } = event.target;
     if (_.isNaN(attemptNumber)) {
-      attemptNumber = '';
-      unlimitedAttempts = true;
+      if (value?.startsWith(defaultValue)) {
+        attemptNumber = value.replace(' (Default)');
+      } else {
+        attemptNumber = '';
+        unlimitedAttempts = true;
+      }
     } else if (attemptNumber < 0) {
       attemptNumber = 0;
     }
     updateSettings({ scoring: { ...scoring, attempts: { number: attemptNumber, unlimited: unlimitedAttempts } } });
+  };
+
+  const handleOnChange = (event) => {
+    let newMaxAttempt = event.target.value;
+    if (newMaxAttempt === defaultValue || newMaxAttempt === '') {
+      newMaxAttempt = `${defaultValue} (Default)`;
+    } else if (newMaxAttempt < 0) {
+      newMaxAttempt = 0;
+    }
+    setLocal(newMaxAttempt);
   };
 
   const handleWeightChange = (event) => {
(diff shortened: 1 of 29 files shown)


src/editors/containers/ProblemEditor/components/EditProblemView/SettingsWidget/hooks.js
```diff
@@ -117,19 +118,35 @@ export const resetCardHooks = (updateSettings) => {
...
 };
 
-export const scoringCardHooks = (scoring, updateSettings) => {
+export const scoringCardHooks = (scoring, updateSettings, defaultValue) => {
+  const [local, setLocal] = module.state.local(scoring.attempts.number);
   const handleMaxAttemptChange = (event) => {
     let unlimitedAttempts = false;
     let attemptNumber = parseInt(event.target.value);
+    const { value } = event.target;
     if (_.isNaN(attemptNumber)) {
```
jesperhodge, line 127: Apparently the attemptNumber used to actually be a number. Now that this has changed because the input type is different, I think we don't want to change the type of the redux state from number to string. So it would be good to change the input from a string into a number before calling `updateSettings`.

src/editors/data/redux/thunkActions/problem.js
```diff
@@ -1,10 +1,13 @@
 import _ from 'lodash-es';
 import { actions } from '..';
+import * as requests from './requests';
 import { OLXParser } from '../../../containers/ProblemEditor/data/OLXParser';
 import { parseSettings } from '../../../containers/ProblemEditor/data/SettingsParser';
 import { ProblemTypeKeys } from '../../constants/problem';
 import ReactStateOLXParser from '../../../containers/ProblemEditor/data/ReactStateOLXParser';
 import { blankProblemOLX } from '../../../containers/ProblemEditor/data/mockData/olxTestData';
+import { camelizeKeys } from '../../../utils';
+import * as module from './problem';
```
jesperhodge, line 10: I generally don't think self-imports are a good idea - they make the application logic more convoluted and add unnecessary complexity, and are not normally used in React applications. Is this really necessary? Why do we need to mock out `loadProblem`, and if we do, why do we not have it in a separate module?

src/editors/data/redux/thunkActions/problem.js
```diff
@@ -43,15 +46,35 @@ export const getDataFromOlx = ({ rawOLX, rawSettings }) => {
...
+      const defaultSettings = {};
+      Object.entries(response.data).forEach(([key, value]) => {
+        if (advancedProblemSettingKeys.includes(key)) {
+          defaultSettings[key] = value.value;
+        }
+      });
+      dispatch(actions.problem.updateField({ defaultSettings: camelizeKeys(defaultSettings) }));
+      module.loadProblem({ rawOLX, rawSettings, defaultSettings })(dispatch);
+    },
+    onFailure: () => { module.loadProblem({ rawOLX, rawSettings, defaultSettings: {} })(dispatch); },
```
jesperhodge, line 70: how about just `loadProblem` without module?

jesperhodge (changes requested):
Really awesome work. Mostly works like a charm, but I found a couple of problems concerning string validation.
Also, if you look into the console after changing these settings, there are a whole bunch of errors. We should ideally fix all of them that are related to this and open issues for the other ones. But to be more pragmatic, it is fine to identify whether there are new errors after this PR or not, and only fixing these, but still open issues for the other ones.

rayzhou-bit (commented):
This looks really good for what we can do. Dream scenario would be that paragon adds a non-editable text or something following the editable field.

rayzhou-bit (commented):
Should "If a value is not set, unlimited attempts are allowed" be removed now?

Seeing "undefined (Default)" looks strange. Maybe it should be "unlimited (Default)" even if checkbox is unchecked?

src/editors/containers/ProblemEditor/index.test.jsx
```diff
@@ -62,17 +63,27 @@ describe('ProblemEditor', () => {
     test('assets loaded, block and studio view not yet loaded, Spinner appears', () => {
       expect(shallow(<ProblemEditor {...props} assetsFinished />)).toMatchSnapshot();
     });
+    test('advanceSettings loaded, block and studio view not yet loaded, Spinner appears', () => {
+      expect(shallow(<ProblemEditor {...props} advanceSettingsFinished />)).toMatchSnapshot();
```
jesperhodge, line 67: Same as the other comment - I would feel a bit more comfortable if we at least tested here that an actual spinner appears (find in shallow output) rather than adding a snapshot.
While we're at it... same problem for the tests above. Since they are basically exactly the same test logic, maybe if you test the behavior in this test, can you just copy it also to the tests above?
By the way I don't have a problem with leaving the snapshots, in principle. My suggestion is just a little test case to add to this.
`const wrapper = shallow(...)`
`expect(wrapper.find(Spinner)).toBe(true)` (I don't know if it returns true or what the return value is, but something like that.)

jesperhodge (changes requested):
Very nice work!! I like the new checkbox and how you fixed the little issues around numeric / string values. I actually didn't quite understand how you managed to prevent the user from editing string content in the input field, the outcome of that is very well-working and impressive. I'd be curious how you managed that if you like to share.
The new test cases are very good, you cover all the potential bugs like string and negative values and make sure everything works perfectly.
I was feeling a bit uncomfortable with the growing amount of conditional cases in the ProblemEditor component and how it can be tested that they work correctly. Maybe the logic would be better placed in the hooks.js files. But we have an issue for testing the lifecycle of that component already, dealing with lifecycle testing and that sort of stuff, so that is out of scope here. It is great that you have tests that cover testing rendering of spinner versus editor content, so I suggested adding more explicit expects to the snapshot tests in one of the comments. The nice thing with that is that with that little change the conditionals logic is sufficiently tested and there really is no need to go into more detail or worry about everything working the way we imagine.

src/editors/data/redux/requests/reducer.js
```diff
@@ -19,6 +19,7 @@ const initialState = {
   [RequestKeys.checkTranscriptsForImport]: { status: RequestStates.inactive },
   [RequestKeys.importTranscript]: { status: RequestStates.inactive },
   [RequestKeys.fetchVideoFeatures]: { status: RequestStates.inactive },
+  [RequestKeys.fetchAdvanceSettings]: { status: RequestStates.inactive },
```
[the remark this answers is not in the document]
jesperhodge: Ironically my comment on your typo also has a typo. ^^

KristinAoki:
> Should "If a value is not set, unlimited attempts are allowed" be removed now?

Jon's new mock for me still included it, so keeping it in unless he explicitly decided otherwise.

> Seeing "undefined (Default)" looks strange. Maybe it should be "unlimited (Default)" even if checkbox is unchecked?

I think it makes since for it to say "null (Default)"  because that is the value that is set in the advanced settings.

jesperhodge (approved):

PR #288, opened 2023-03-23
feat: error on funky multiple choice tags.

connorhaugh:
This change increases the frequency at which odd tags (like shuffle) on multiple choice problems don't go to the visual editor.
Instead, they will now divert to the advanced editor. There might be other places where we need to follow a similar pattern, but we don't know those yet.

Testing instructions:
- Follow the process of creating a shuffle problem in advanced editor using this olx: (edit this Q to get the xml)

```
<problem>
  <multiplechoiceresponse>
    <label>What Apple device competed with the portable CD player?</label>
    <choicegroup type="MultipleChoice" shuffle="true">
      <choice correct="false">The iPad</choice>
      <choice correct="false">Napster</choice>
      <choice correct="true">The iPod</choice>
      <choice correct="false">The vegetable peeler</choice>
    </choicegroup>
  </multiplechoiceresponse>
</problem>

```

It should save, and open again in the advanced editor.

In addition, all the other, normal, uses of problems should still work.

### DIFF
--- a/src/editors/containers/ProblemEditor/data/OLXParser.js
+++ b/src/editors/containers/ProblemEditor/data/OLXParser.js
@@ -72,6 +72,10 @@ export class OLXParser {
     const answers = [];
     let data = {};
     const widget = _.get(this.problem, `${problemType}.${widgetName}`);
+    const permissableTags = ['choice', '@_type', 'compoundhint', 'option', '#text'];
+    if (_.keys(widget).some((tag) => !permissableTags.includes(tag))) {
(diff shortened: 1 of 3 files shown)


rayzhou-bit (commented):
This looks good to me. Quick and smart way to see if it should go to advanced.
My devstack hasn't been working and I haven't gotten around to getting on hosted again so I wasn't able to test it. I think somebody else should test before merge.

KristinAoki (approved):

PR #294, opened 2023-04-05
feat: default scoring to 1, not zero.

connorhaugh:
Since every course has graded problems, every course author will be editing or creating problems, and every learner will need to receive points for any problem in a graded assignment in order to earn a passing grade. So if what I encountered and am reporting here is reproducible, this should probably be addressed before the problem editor feature is released.


jristau1984 (approved):
lgtm.

PR #295, opened 2023-04-05
feat: add alert to notify video id changes

KristinAoki:
JIRA Ticket: [TNL-10297](https://2u-internal.atlassian.net/browse/TNL-10297)

When a user changes the video id and does not have a fallback video or a url video source, the backend creates a link to the edx hosted video. This is inserted when the video block is loaded after the previous save. However, when the video id for the block is later changed, the auto-generated link does not update. As a result, the video block still points to the old video, not the one referenced by the video id. This PR adds an alert to bring this behavior to the attention of users.

### DIFF
--- a/src/editors/containers/VideoEditor/components/VideoSettingsModal/components/VideoSourceWidget/hooks.test.jsx
+++ b/src/editors/containers/VideoEditor/components/VideoSettingsModal/components/VideoSourceWidget/hooks.test.jsx
@@ -1,7 +1,15 @@
 import { dispatch } from 'react-redux';
 import { actions } from '../../../../../../data/redux';
+import { MockUseState } from '../../../../../../../testUtils';
 import * as requests from '../../../../../../data/redux/thunkActions/requests';
 import * as hooks from './hooks';
+
+jest.mock('react', () => ({
+  ...jest.requireActual('react'),
+  useRef: jest.fn(val => ({ current: val })),
+  useEffect: jest.fn(),
+  useCallback: (cb, prereqs) => ({ cb, prereqs }),
+}));
 
 jest.mock('react-redux', () => {
   const dispatchFn = jest.fn();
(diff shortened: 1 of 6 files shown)


src/editors/containers/VideoEditor/components/VideoSettingsModal/components/VideoSourceWidget/index.jsx
```diff
@@ -52,7 +52,12 @@ export const VideoSourceWidget = ({
       [widgetHooks.selectorKeys.allowVideoSharing]: widgetHooks.genericWidget,
     },
   });
-  const { updateVideoId, updateVideoURL } = hooks.sourceHooks({ dispatch });
+  const { videoIdChangeAlert } = hooks.videoIdChangeAlert();
+  const { updateVideoId, updateVideoURL } = hooks.sourceHooks({
+    dispatch,
+    previousVideoId: videoId.formValue,
+    videoIdChangeAlert,
```
rayzhou-bit, line 59: nit: what do you think about something like this to better show what is being used by the hook?
```setAlert: videoIdChangeAlert.set```

File: src/editors/containers/VideoEditor/components/VideoSettingsModal/components/VideoSourceWidget/hooks.test.jsx [hunk 2]
```diff
@@ -120,4 +152,23 @@ describe('VideoEditorHandout hooks', () => {
...
     });
   });
+  describe('videoIdChangeAlert', () => {
+    beforeEach(() => {
+      state.mock();
+    });
+    afterEach(() => {
+      state.restore();
+    });
+    test('showVideoChangeAlert: state values', () => {
```
rayzhou-bit, line 162: This should be `showVideoIdChangeAlert`. (same for the other 2 below)

rayzhou-bit (changes requested):
This looks great! Just have a test name update you should make to avoid future confusion. Also 1 nit suggestion.

rayzhou-bit (approved):
lgtm!

PR #297, opened 2023-04-05
feat: revert group feedback expanded text field

rayzhou-bit:
This PR reverts the expanded text area changes applied for the Group Feedback widget. This is a proposed change to the Group Feedback not saving properly for the simple problem editor.

https://2u-internal.atlassian.net/browse/TNL-10604


connorhaugh (approved):
LGTM!

PR #298, opened 2023-04-06
feat: add check to sanitize answer ranges

KristinAoki:
JIRA Ticket: [TNL-10311](https://2u-internal.atlassian.net/browse/TNL-10311)

This PR adds a check to see if a numerical response's answer range is a valid format ([smaller, larger]). Currently any order is valid for an answer range, such as (larger, smaller]. We need to enforce the expected format to guarantee answers are graded properly.


rayzhou-bit (changes requested):
We would want these to work for fractions as well as the LMS does take fraction values.

rayzhou-bit (approved):
lgtm!

PR #299, opened 2023-04-06
fix  image resize

jesperhodge:
JIRA link: https://2u-internal.atlassian.net/browse/TNL-10584

### Description:

This is a bug where the image resizing in text editor and problem editor was completely broken. Putting in a text value when the aspect ratio lock was enabled would change both values but not to the size you wanted. If you disabled the lock, not just one but both values would change.

This is a problem that mostly affects images that are rectangular, not square. There's an example image below which is one that caused problems on prod.

### Main fixes:

- when I keep the image ratio locked, I can change one value (like width) and the other will jump to the proportionate value, but rounded to full pixels.
- when I unlock the aspect ratio and change a value, then click save on the image dimension modal, only the one value will change, which will stretch the image in whatever direction. This is reflected in the tinymce image and then the updated value will appear when I reopen the image dimension modal. It is not possible to reset the image to the original dimensions any longer. The new values are saved.
- The image dimensions in the edit image settings modal should always reflect the actual dimensions of the image when I look at it e.g. in the course outline. (Otherwise I may click save and the image is squished.)
- There was a problem with deselecting an image: when you edit image dimensions and then save or press cancel, the "edit image" button will not disappear, but the image is not selected anymore. When you do not click anything else but immediately click on this button, sometimes (at least the second or third time you do this) this will throw an error. I fixed it so it will just open the default "select image" modal.

### Other requirements:

- Resizing the image means that when I open the dimensions update, I see the new dimensions.
- Images in the editor are now displayed with the correct dimensions, proportional or stretched, if those dimensions don't exceed the size of the editor.
- A known smaller bug emerging from this is that when you have more than one instance of the same image in the same editor, you cannot get or set its dimensions correctly. I believe I have gotten it into the following state: When you click one of the copies, you will either get the correct dimensions of the selected copy, or if not, it will display the original image dimensions. When you edit the dimensions, the correct copy of the image will be updated.

### Out of Scope:

- This cannot handle more than one instance of the same image properly. There will be a separate bug issue for this.
- Sometimes, when you edit image dimensions and then reopen the image dimension modal, the dimensions will be null and thus just not appear in the modal - randomly. This is a bug as well.

### How to test:

- Go to the ProblemEditor or TextEditor
- click the upload image button in the Question / Text editor
- upload this:

resizeTestImage

- click to open resize modal
- with lock enabled, change one value. Check that only the other value changes to keep the aspect ratio.
- save and reopen. Values should be persisted.
- with lock disabled, change one value. Check that the other value does not change.
- save and reopen. New values should be there, not the old values (which was a bug we fixed).
- Put in dimensions that exceed the width of your screen. When you save, the image in the editor should be scaled down but stay proportional; the image that is saved (check course outline) should have the actual bigger width you gave; when you save and reopen the image dimensions modal, you should get the actual dimensions, not the smaller dimensions from the editor view.
- Resize the image by clicking and dragging. When you open the image dimensions modal, the dimensions should have changed accordingly.
- Edit image dimensions in the modal and then close by saving. You probably see the button for editing the image is still present, but the image doesn't have a blue border, so it isn't selected. The very next thing to click on after closing the editor is the button. This should not throw an error but just open the "select image" modal. Repeat this two more times to make sure there is no error.
- Upload the same image a second time. When editing the image, make sure that whatever copy of the image you edit, it changes the dimensions of the correct copy. You will notice that sometimes when you open the edit image modal it will not show the correct dimensions but the original image dimensions. That is an expected bug with this fix.


src/editors/sharedComponents/ImageUploadModal/ImageSettingsModal/hooks.js
```diff
@@ -12,7 +12,7 @@ export const state = {
   isDecorative: (val) => React.useState(val),
   isLocked: (val) => React.useState(val),
   local: (val) => React.useState(val),
-  lockDims: (val) => React.useState(val),
+  lockAspectRatio: (val) => React.useState(val),
```
rayzhou-bit, line 15: Nice rename

rayzhou-bit (approved):
lgtm! Tests look great! I would like somebody else to review as well since there's tinyMCE jank.

jesperhodge:
Needs a fix with two additions defined by Jonathan:
- add a text link to reset image (only when proportions have changed)
- fix problem that image shrinks to TinyMCE window

rayzhou-bit (approved):
lgtm! Just a note that the alt text seems to have a similar issue as dimensions. It isn't consistent but sometimes when there are duplicates of the same image, the alt text for the other image will show. This goes away when you re-open more and more. I expect it would be fixed with the same solution when the dimension for duplicate images are fixed.

jesperhodge:
> lgtm! Just a note that the alt text seems to have a similar issue as dimensions. It isn't consistent but sometimes when there are duplicates of the same image, the alt text for the other image will show. This goes away when you re-open more and more. I expect it would be fixed with the same solution when the dimension for duplicate images are fixed.

Yeah, we may just need to do a followup issue on this at some point. It's a very difficult problem to solve.

PR #304, opened 2023-04-12
feat: XMLBuilder and XMLParser performing unwanted processing in encoding / parsing 

rayzhou-bit:
When the simple problem editor loads OLX, the XMLBuilder encodes the OLX before tinyMCE decodes it again. This means that if the block contains any encoded HTML entities, they will not appear decoded on the editor.
LMS simply decodes the OLX. The simple editor should match that behavior.

An example of what is happening currently:
LMS: ` `
Block OLX: `&#160;`
After XMLBuilder processing: `&amp;#160;`
After tinyMCE processing: `&#160;`  (this is what is seen in the simple problem editor)

This PR keeps XMLBuilder from encoding the OLX.

https://2u-internal.atlassian.net/browse/TNL-10627

=========================

Somewhat related, the XMLParser automatically parses numbers like 0x12 as hexes. This PR will turn this parsing off as well.

https://2u-internal.atlassian.net/browse/TNL-10629

### DIFF
--- a/src/editors/containers/ProblemEditor/data/OLXParser.js
+++ b/src/editors/containers/ProblemEditor/data/OLXParser.js
@@ -481,7 +481,6 @@ export class OLXParser {
   }
 
   getParsedOLXData() {
-    console.log('test', this.parsedOLX)
     if (_.isEmpty(this.problem)) {
       return {};
     }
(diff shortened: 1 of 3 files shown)


KristinAoki (changes requested):
Can you add a test that confirms the parser return the expected values

PR #306, opened 2023-04-14
feat: add xml linter to code mirror

KristinAoki:
JIRA Ticket: [TNL-10609](https://2u-internal.atlassian.net/browse/TNL-10609)

When a course author is using the advanced problem editor and they have malformed xlm, there is a toast message that appears notifying them that their changes could not be saved and to try again later. This message does not clearly communicate to the author the actual error in the xlm. The PR adds a xlm linter plugin to code mirror. When the xlm is malformed, such as unclosed tags, a red squiggle will appear at the location of the error. On hover a message describing the error will appear.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,6 +11,7 @@
       "dependencies": {
         "@codemirror/lang-html": "^6.0.0",
         "@codemirror/lang-xml": "^6.0.0",
+        "@codemirror/lint": "^6.2.1",
         "@codemirror/state": "^6.0.0",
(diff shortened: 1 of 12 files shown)


connorhaugh (approved):
> How are we handling errors that are specific to the OLX, not XML? It seems like the save might fail, even though the linter says nothing. What messaging could we provide?

After further investigation, it seems that OLX-specfic errors occur at render time, not at save, so it is safe to have this for now.

PR #310, opened 2023-04-19
feat: add no answer confirmation alert

KristinAoki:
JIRA Ticket: [TNL-10312](https://2u-internal.atlassian.net/browse/TNL-10312)

This PR adds a confirmation alert when a problem does not have a correct answer selected or the correct answer selected has no title. When the user clicks save an alert will appear and if the user confirms the save, the problem will save. Otherwise the alert will disappear and the user will be able to edit the problem.


rayzhou-bit:
I'm not seeing the pop up when you select an empty option as the correct answer. Should there be a pop up for this?

KristinAoki:
> I'm not seeing the pop up when you select an empty option as the correct answer. Should there be a pop up for this?

yes, thank you for catching this! I forgot to test with multiple answers.

PR #311, opened 2023-04-20
feat: tinyMCE formatting and parsing

rayzhou-bit:
Blocks that look like this:
image

are turning into this when just opening and saving the block without changes:
image

The weird spacing and formatting is due to how studio displays text wrapped in a `<p>` block tag. Before the visual problem editor update, this text was not wrapped in `<p>`. However, tinyMCE requires text to be wrapped in block tags.

This PR fixes this by doing the following:
1. Have tinyMCE default to a `<div>` block tag instead of a `<p>` block tag. Allow selection of div block tag.
2. Strip the block tag around `<label>` tags.

Internal ticket: https://2u-internal.atlassian.net/browse/TNL-10632

### DIFF
--- a/src/editors/containers/ProblemEditor/data/ReactStateOLXParser.js
+++ b/src/editors/containers/ProblemEditor/data/ReactStateOLXParser.js
@@ -188,6 +188,22 @@ class ReactStateOLXParser {
   addQuestion() {
     const { question } = this.editorObject;
     const questionObject = this.questionParser.parse(question);
+    /* Removes block tags like <p> or <h1> that surround the <label> format.
+      Block tags are required by tinyMCE but have adverse effect on css in studio.
+      */
+    questionObject.forEach((tag, ind) => {
+      const tagName = Object.keys(tag)[0];
+      let label = null;
+      tag[tagName].forEach(subTag => {
+        const subTagName = Object.keys(subTag)[0];
+        if (subTagName === 'label') {
+          label = subTag;
+        }
+      });
+      if (label) {
+        questionObject[ind] = label;
+      }
+    });
     return questionObject;
   }
 
--- a/src/editors/sharedComponents/TinyMceWidget/pluginConfig.js
+++ b/src/editors/sharedComponents/TinyMceWidget/pluginConfig.js
@@ -14,7 +14,7 @@ const pluginConfig = ({ isLibrary, placeholder, editorType }) => {
   const quickToolbar = editorType === 'expandable' ? plugins.quickbars : '';
   const inline = editorType === 'expandable';
   const toolbar = editorType !== 'expandable';
-  const defaultFormat = editorType === 'text' ? 'p' : 'div';
+  const defaultFormat = (editorType === 'question' || editorType === 'expandable') ? 'div' : 'p';
 
   return (
     StrictDict({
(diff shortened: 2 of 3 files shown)


src/editors/containers/ProblemEditor/data/ReactStateOLXParser.js
```diff
@@ -188,6 +188,22 @@ class ReactStateOLXParser {
   addQuestion() {
     const { question } = this.editorObject;
     const questionObject = this.questionParser.parse(question);
+    /* Removes block tags like <p> or <h1> that surround the <label> format.
+      Block tags are required by tinyMCE but have adverse effect on css in studio.
+      */
+    questionObject.forEach((tag, ind) => {
```
connorhaugh, line 194: Will we have to do this in the explanation feild as well for any tags?

connorhaugh (commented):
> block tag instead of a
block tag

just so we are clear, we are replacing block tags with divs or nothing right?

src/editors/containers/ProblemEditor/data/ReactStateOLXParser.js
KristinAoki: I don't think so. There is no special formatting for the explanation field.

rayzhou-bit:
> > block tag instead of a
> > block tag
> 
> just so we are clear, we are replacing block tags with divs or nothing right?

I couldn't get tinyMCE to be ok with not having block tags at all. This change has tinyMCE insert div tags instead of p tags when there's no block tags (and this doesn't seem to have that extra bottom padding when saved to studio). If there is an existing block tag, tinyMCE should load with that block tag.

... I just realized I completely butchered my description because I didn't have ```` around the tags.

src/editors/containers/ProblemEditor/data/ReactStateOLXParser.js
rayzhou-bit: From read the docs, explanation fields has had `<p>` tags baked in in the OLX so I think this issue is more with question and answer fields. 

There's an example of explanation OLX here (from before my RtDoc edits so I'm not just referencing myself):
https://edx.readthedocs.io/projects/edx-partner-course-staff/en/latest/exercises_tools/single_select.html#answer-pools

src/editors/sharedComponents/TinyMceWidget/pluginConfig.js
```diff
@@ -14,6 +14,7 @@ const pluginConfig = ({ isLibrary, placeholder, editorType }) => {
   const quickToolbar = editorType === 'expandable' ? plugins.quickbars : '';
   const inline = editorType === 'expandable';
   const toolbar = editorType !== 'expandable';
+  const defaultFormat = editorType === 'text' ? 'p' : 'div';
```
KristinAoki, line 17: This causes errors for the the explanation field. The current code expects the explanation field to be wrapped in paragraph tags to properly save and add the explanation header. When the block uses `div` tags the explanation is not added.

src/editors/sharedComponents/TinyMceWidget/pluginConfig.js
```diff
@@ -93,6 +94,8 @@ const pluginConfig = ({ isLibrary, placeholder, editorType }) => {
         convert_urls: false,
         placeholder,
         inline,
+        block_formats: 'Div=div;Paragraph=p;Header 1=h1;Header 2=h2;Header 3=h3;Header 4=h4;Header 5=h5;Header 6=h6;Preformatted=pre',
```
KristinAoki, line 97: Nit: reorder this so it goes headers, div, paragraph, preformatted.

KristinAoki (changes requested):

KristinAoki:
> This PR fixes this by doing the following:
> 1. Have tinyMCE default to a <div> block tag instead of a <p> block tag. Allow selection of div block tag.


Do you think it makes sense for this to only be true for complex text areas like answers, feedback, hint, etc.? This would help resolve the explanation error. The issue with the question widgets labels are resolved with your parser update.

rayzhou-bit:
> Do you think it makes sense for this to only be true for complex text areas like answers, feedback, hint, etc.? This would help resolve the explanation error. The issue with the question widgets labels are resolved with your parser update.

Yeah. Question, answer, feedback, hint should use div default while explanation and text (text editor) use p default. Good catch on the explanation field.

I think we didn't realize just how closely we needed to stick to example OLX's to keep formatting in tact.

KristinAoki (approved):

PR #314, opened 2023-04-21
feat: remove layered button

connorhaugh:
Having a button inside a button makes the button content focusable, not just the button, allowing users to click the wrong thing. This removes that but keeps the styling.

Screenshot 2023-04-21 at 9 56 44 AM


kenclary (approved):
good catch

PR #315, opened 2023-04-21
fix: advanced editor not saving

KristinAoki:
JIRA Ticket: [TNL-10647](https://2u-internal.atlassian.net/browse/TNL-10647)

The addition of the no answer alert in the visual editor created a bug in the advanced editor that prevented the editor from being saved. The editor was trying to find answers that did not exist. This PR adds a check if the problem editor is in the advanced view. If it is in the advanced editor, it will skip over the no answer check.


connorhaugh (approved):
Thanks for adding the tests too!

PR #317, opened 2023-04-25
feat: default randomization to never

rayzhou-bit:
In the problem editor, the randomization setting is typically used for randomization with a python script. However, the setting is also needed for the shuffle answer functionality. This PR ensures that the randomization setting can be configured for all problems and defaults the setting to 'never'.

Jira: https://2u-internal.atlassian.net/browse/TNL-10652


kenclary (approved):
good catch

PR #319, opened 2023-04-26
feat: remove custom css with paragon upgrade

KristinAoki:
JIRA Ticket: [TNL-10618](https://2u-internal.atlassian.net/browse/TNL-10618)

Removes custom css for green checkboxes/radios when selected. This functionality is now supported by Paragon.


connorhaugh (approved):
Great -- can confirm this was added in https://github.com/openedx/paragon/releases/tag/v20.31.1

PR #322, opened 2023-04-28
feat: add a button to return to studio

KristinAoki:
JIRA Ticket: [TNL-10387](https://2u-internal.atlassian.net/browse/TNL-10387)

### DIFF
--- a/src/editors/VideoSelectorPage.jsx
+++ b/src/editors/VideoSelectorPage.jsx
@@ -2,7 +2,7 @@ import React from 'react';
 import { Provider } from 'react-redux';
 import PropTypes from 'prop-types';
 import ErrorBoundary from './sharedComponents/ErrorBoundary';
-import { VideoSelector } from './VideoSelector';
+import VideoSelector from './VideoSelector';
 import store from './data/store';
 
 const VideoSelectorPage = ({
@@ -10,8 +10,9 @@ const VideoSelectorPage = ({
   lmsEndpointUrl,
   studioEndpointUrl,
 }) => (
-  <ErrorBoundary>
     <Provider store={store}>
+        <ErrorBoundary {...{ courseId, studioEndpointUrl }}>
+
       <VideoSelector
         {...{
           learningContextId: courseId,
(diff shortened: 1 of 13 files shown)


src/editors/VideoSelectorPage.jsx
```diff
@@ -2,25 +2,25 @@ import React from 'react';
...
 const VideoSelectorPage = ({
   courseId,
   lmsEndpointUrl,
   studioEndpointUrl,
 }) => (
-  <ErrorBoundary>
-    <Provider store={store}>
+  <Provider store={store}>
+    <ErrorBoundary {...{ courseId, studioEndpointUrl }}>
       <VideoSelector
```
connorhaugh, line 15: Unrelated to your fix: this kind of thing breaks the abstraction of editors..... :( Why didn't they follow the pattern from the problem editor?

src/editors/sharedComponents/ErrorBoundary/ErrorPage.jsx
```diff
@@ -1,48 +1,88 @@
...
-        )}
-        <Button onClick={global.location.reload()}>
-          {intl.formatMessage(messages.unexpectedErrorButtonLabel)}
-        </Button>
-      </Col>
-    </Row>
-  </Container>
-);
+}) => {
+  const courseOutlineUrl = `${studioEndpointUrl}/course/${courseId}`;
```
connorhaugh, line 26: what happens when we are editing a library? probably better to use terms like `learningContextId` and things like that.

connorhaugh (changes requested):
This should be learning-context agnostic as best as possible.

src/editors/sharedComponents/ErrorBoundary/ErrorPage.jsx
connorhaugh: The problems here are twofold:
1. we should rename so it is learning context agnostic.
2. We should make sure returning to the "course" page works for a library

connorhaugh (approved):
Sweet-- thanks for making those changes.

PR #323, opened 2023-05-02
feat: add social share widget

KristinAoki:
JIRA Ticket: [AU-1188](https://2u-internal.atlassian.net/browse/AU-1188)

This PR moves the checkbox that allows public access to video into its own widget/section. The section is not rendered if the waffle flag `video_config.public_video_share` is disabled. The checkbox is disabled when the setting is set at the course level and enabled when the setting is set on video-by-video basis.

This PR is dependent on [edx-platform PR 32157](https://github.com/openedx/edx-platform/pull/32157).


leangseu-edx (approved):
LGTM

PR #325, opened 2023-05-04
fix: expandable textarea should save on blur

rayzhou-bit:
The expandable textareas employing tinyMCE in problem editor answer fields were only saving to react state when the problem was saved or the problem type was changed. This causes an issue when deleting answer fields that have not been saved to react state. This PR ensures that we save these tinyMCE fields on blur.

https://2u-internal.atlassian.net/browse/TNL-10644


connorhaugh (approved):
LGTM

PR #327, opened 2023-05-05
fix: show setting note for course set social share

KristinAoki:
When a course has 'all-on' or 'all-off' for the social sharing in the course outline, there should be two messages under the checkbox notifying the user that the setting overridden by the course settings and that they can be changed on the course outline page. When a course has 'per-video' for the social sharing in the course outline, there will be no messages bewlow the checkbox.


schenedx (approved):
👍

PR #328, opened 2023-05-07
fix: delete answers without changing expandable

rayzhou-bit:
[A previous PR](https://github.com/openedx/frontend-lib-content-components/pull/325) introduces an issue with expandable textarea where images can not be inserted anywhere except at the beginning of the textarea.

This PR fixes this by approaching the solution from a different way. The answer fields are saved to the state before deletion, allowing for the correct answer to be deleted.

This PR also allows the last answer to be deleted by wiping the values of the answer, selectedFeedback and unselectedFeedback text fields for this last answer when delete is clicked.

https://2u-internal.atlassian.net/browse/TNL-10644

### DIFF
--- a/src/editors/containers/ProblemEditor/components/EditProblemView/AnswerWidget/hooks.js
+++ b/src/editors/containers/ProblemEditor/components/EditProblemView/AnswerWidget/hooks.js
@@ -16,13 +16,30 @@ export const removeAnswer = ({
   dispatch,
 }) => () => {
   if (RichTextProblems.includes(problemType)) {
+    let newAnswers = [];
     const currentAnswerTitles = fetchEditorContent({ format: '' }).answers;
     answers.forEach(ans => {
-      dispatch(actions.problem.updateAnswer({
-        ...ans,
-        title: currentAnswerTitles?.[ans.id] || ans.title,
-        correct: ans.correct,
-      }));
+      if (ans !== answer) {
+        newAnswers = [
+          ...newAnswers,
+          currentAnswerTitles?.[ans.id] || '',
+        ];
+      }
+    });
+    const editorObject = { hints: [] };
+    const EditorsArray = window.tinymce.editors;
+    let iter = 0;
+    Object.entries(EditorsArray).forEach(([id, editor]) => {
+      if (Number.isNaN(parseInt(id))) {
+        if (id.startsWith('answer')) {
+          const { editorAnswers } = editorObject;
+          const answerId = id.substring(id.indexOf('-') + 1);
+          if (iter < newAnswers.length) {
+            editorObject.editorAnswers = { ...editorAnswers, [answerId]: editor.setContent(newAnswers[iter]) };
+            iter++;
+          }
+        }
+      }
     });
   }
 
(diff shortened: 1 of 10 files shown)


src/editors/containers/ProblemEditor/components/EditProblemView/AnswerWidget/hooks.js
```diff
@@ -9,7 +9,25 @@ export const state = StrictDict({
...
 });
 
-export const removeAnswer = ({ answer, dispatch }) => () => {
+export const removeAnswer = ({
+  answers,
+  answer,
+  problemType,
+  dispatch,
+}) => () => {
+  const richTextProblems = [ProblemTypeKeys.SINGLESELECT, ProblemTypeKeys.MULTISELECT];
```
connorhaugh, line 18: It's probably good to define this constant somewhere central, so we don't break a bunch of things when we change which problem types use expanding text areas .

connorhaugh:
Question: why have we not removed the .onBlur() in this pr, if we are taking this new direction to fix the bug?

rayzhou-bit:
> Question: why have we not removed the .onBlur() in this pr, if we are taking this new direction to fix the bug?

It is removed in src/editors/sharedComponents/TinyMceWidget/hooks.js

src/editors/containers/ProblemEditor/components/EditProblemView/AnswerWidget/hooks.js
rayzhou-bit: That's a good idea

connorhaugh:
> > Question: why have we not removed the .onBlur() in this pr, if we are taking this new direction to fix the bug?
> 
> It is removed in src/editors/sharedComponents/TinyMceWidget/hooks.js

Lol bad reading comprehension on my part. Looks good.

connorhaugh (approved):

src/editors/containers/ProblemEditor/components/EditProblemView/AnswerWidget/hooks.js
```diff
@@ -2,14 +2,30 @@ import { useState, useEffect } from 'react';
...
 
-export const removeAnswer = ({ answer, dispatch }) => () => {
+export const removeAnswer = ({
+  answers,
+  answer,
+  problemType,
+  dispatch,
+}) => () => {
+  if (RichTextProblems.includes(problemType)) {
+    const currentAnswerTitles = fetchEditorContent({ format: 'text' }).answers;
```
KristinAoki, line 19:
```suggestion
    const currentAnswerTitles = fetchEditorContent({ format: '' }).answers;
```

KristinAoki: giving it the parameter text converts it from rich text to plain text. when saving it to the redux we should maintain the rich text

KristinAoki (changes requested):

KristinAoki (commented):
This seems to be working as expected. However, there is one small bug that I have observed. On the first delete of an answer, where there are three answers and the last one is empty, both fields become empty after the delete occurs. I am only able to reproduce the error on the initial load of the answers. Everything is good after the first delete has been made.

rayzhou-bit:
I'm realizing feedback is not deleted properly in stage either so I'm coding to account for this as well.

KristinAoki (changes requested):
Feedback is properly maintained when answers are deleted from the bottom up. However, if an answer that has answers below it (ex: answer B with answer C and D below) is deleted the feedback is cleared from the following answers. The section also automatically collapses, which is not the expected behavior.

PR #330, opened 2023-05-10
chore: refactor OLXParser and related doc strings

KristinAoki:
JIRA Ticket: [TNL-10417](https://2u-internal.atlassian.net/browse/TNL-10417)


src/editors/containers/ProblemEditor/data/OLXParser.test.js
```diff
@@ -23,227 +23,275 @@ import {
...
-    const question = olxparser.parseQuestions(problemType);
-    expect(question).toBe(blankQuestionOLX.question);
+  describe('parseMultipleChoiceAnswers()', () => {
+    describe('given a problem with no answers', () => {
+      const { answers } = multipleChoiceWithoutAnswersOlxParser.parseMultipleChoiceAnswers(
+        'multiplechoiceresponse',
+        'choicegroup',
+        'choice',
+      );
+      it('should return an array of objects with length one', () => {
```
jesperhodge, line 177: Nit: this is really a nit, you don't need to fix this. just an observation.
this is just a feeling about the wording here. If I ask myself what we are testing, it is probably that the answers are correct, or that all the answers are extracted. So while testing for length 1 is important, probably the name of the test should describe more what we are testing for (what we expect). I may be scratching my head here: "If there are no multiple choice answers in the OLX, why is an answer returned?" so the test name could be for example, "should return a default answer". And this is just for the naming of the test, the test itself is great.

PR #334, opened 2023-05-19
fix: feedback clearing on type change

KristinAoki:
When a user changed from a single select or multi-select problem type to another, the feedback was cleared out. This is because the data wasn't stored in the redux store before the problem type was updated. Now when changing between problem types, the feedback is saved to the redux store.


connorhaugh (approved):
Steps I took to test this:
1. checkout PR
2. make a multi select problem
3. Add some feedback to  an answer
4. Switch to single select
5. See that the feedback stays
6. Switch back to single
7. Feedback stays

PR #335, opened 2023-05-23
feat: replace canvas with jest-canvas-mock

KristinAoki:
The `canvas` package breaks all the tests when upgrading to node 18. This PR replaces it with `jest-canvas-module` package and adds it as dev dependency.


connorhaugh:
For those playing along at home, this package:
- Has significant usage
- Is on an MIT license 
- Is only used in tests.

Therefore, it makes sense to add it.

connorhaugh (approved):

PR #336, opened 2023-05-25
fix: olx parsers with formatted text

KristinAoki:
JIRA Ticket: [TNL-10689](https://2u-internal.atlassian.net/browse/TNL-10689)

Currently if a user formats their text (italicize, bold, etc.) or adds an image all inside the same parent tag, e.g. `<p>`, the order of the text/image will be changed on save and re-open. This causes user to be confused and not trust the visual editor as it is not saving their work as expected. This PR uses `fast-xml-parser`'s attribute `preserve-order` to keep the text in the same order while parsing and building. This attribute is currently used for the question parser, and is now being applied to all places where tinymce is used.

### DIFF
--- a/src/editors/containers/ProblemEditor/data/OLXParser.js
+++ b/src/editors/containers/ProblemEditor/data/OLXParser.js
@@ -101,8 +101,45 @@ export class OLXParser {
     this.parsedQuestionOLX = questionParser.parse(olxString);
     if (_.has(this.parsedOLX, 'problem')) {
       this.problem = this.parsedOLX.problem;
-      this.questionData = this.parsedQuestionOLX[0].problem;
+      this.questionData = this.richTextOlx[0].problem;
+      this.richTextProblem = this.richTextOlx[0].problem;
     }
+  }
+
+  getPreservedAnswersAndFeedback(problemType, widgetName, option) {
+    const preservedProblem = this.richTextProblem;
+    const isChoiceProblem = problemType !== ProblemTypeKeys.NUMERIC ? problemType !== ProblemTypeKeys.TEXTINPUT : false;
+    const preservedAnswers = [];
+    let correctFeedbackTag = option;
+    let incorrectFeedbackTag;
+    if (problemType === ProblemTypeKeys.TEXTINPUT) {
+      [correctFeedbackTag, incorrectFeedbackTag] = option;
+    }
+    preservedProblem.forEach(wrapperTag => {
+      const wrapperTagKeys = Object.keys(wrapperTag);
+      if (wrapperTagKeys.includes(problemType)) {
+        const problemTag = wrapperTag[problemType];
+        problemTag.forEach(obj => {
+          const objKeys = Object.keys(obj);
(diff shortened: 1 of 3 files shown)


src/editors/containers/ProblemEditor/data/OLXParser.js
```diff
@@ -91,18 +92,74 @@ export class OLXParser {
...
+   * olx for the given problem type and widget is parsed. Depending on the
+   * @param {string} problemType - string of the olx problem type
+   * @param {string} widgetName - string of the wrapping tag name
+   *                              (optioninput, choicegroup, checkboxgroup, addition_answer)
+   * @param {string} option - string of the type of answers (choice, option)
+   * @return {array} array containing answer objects and possibly an array of grouped feedback
+   */
+  getPreservedAnswersAndFeedback(problemType, widgetName, option) {
+    const preservedProblem = this.richTextProblem;
+    const isChoiceProblem = problemType !== ProblemTypeKeys.NUMERIC ? problemType !== ProblemTypeKeys.TEXTINPUT : false;
```
jesperhodge, line 132: This is hard to understand. I needed a while to work through all the negations to figure out what value is actually returned. I played around a bit, this seems a more readable version I came up with:
```
const isChoiceProblem = !([ProblemTypeKeys.NUMERIC, ProblemTypeKeys.TEXTINPUT].includes(problemType))
```

src/editors/containers/ProblemEditor/data/OLXParser.js
```diff
@@ -91,18 +92,74 @@ export class OLXParser {
...
+    const preservedAnswers = [];
+    let correctFeedbackTag = option;
+    let incorrectFeedbackTag;
+    if (problemType === ProblemTypeKeys.TEXTINPUT) {
+      [correctFeedbackTag, incorrectFeedbackTag] = option;
     }
+    preservedProblem.forEach(wrapperTag => {
+      const wrapperTagKeys = Object.keys(wrapperTag);
+      if (wrapperTagKeys.includes(problemType)) {
```
jesperhodge, line 141: This function has a very high complexity. There are multiple nested if statements and for loops and it is not possible for me to follow along with the code without spending a large amount of time analyzing it. It seems to me to make this more readable and maintainable, it is helpful to split this up into several well-named functions and variables, and avoid nested ifs / for loops.

jesperhodge: I will mark this as optional. You can decide whether you want to do this here or not. I would of course welcome any improvements of readability, but the parser is already very complex anyway and we can also leave refactoring for another time. With the great amount of work you did in this issue I'm sure you'll be glad to get this merged.

jesperhodge (approved):
This is awesome work! Works like a charm.

connorhaugh (approved):
A lot of code here that _looks_ ok, but I quickly smoke tested by:
Adding some text
adding an image as a child of the P tag. 
Bolding the text inside the p tag
saving
reopening
and the order was preserved and the item stayed bold.

PR #337, opened 2023-05-31
fix: choice parsing for answers with multi lines

KristinAoki:
Follow-up PR to #336.

During testing noticed that problem types that have choice answers did not fully load answers that were multi-lined. Now the full answer loads. The parser was not properly deconstructing the answers for the title and feedback.


kenclary (approved):
lgtm

PR #341, opened 2023-06-07
feat: add returnUrl prop to editors

KristinAoki:
This PR adds a new prop, `returnFunction` to editors. This new prop is necessary for pages that use an editor, but are not connected to a Studio Unit page. This prop is used when exiting an editor (save or cancel).

### DIFF
--- a/src/editors/containers/EditorContainer/hooks.js
+++ b/src/editors/containers/EditorContainer/hooks.js
@@ -23,12 +23,9 @@ export const handleSaveClicked = ({
   dispatch,
   getContent,
   validateEntry,
-  returnUrl,
+  returnFunction,
 }) => {
-  let destination = useSelector(selectors.app.returnUrl);
-  if (returnUrl) {
-    destination = returnUrl;
-  }
+  const destination = useSelector(selectors.app.returnUrl);
   const analytics = useSelector(selectors.app.analytics);
 
   return () => saveBlock({
@@ -36,6 +33,7 @@ export const handleSaveClicked = ({
     content: getContent({ dispatch }),
     destination,
     dispatch,
+    returnFunction,
     validateEntry,
   });
 };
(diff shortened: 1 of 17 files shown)


src/editors/containers/EditorContainer/hooks.js
```diff
@@ -41,12 +49,17 @@ export const cancelConfirmModalToggle = () => {
   };
 };
 
-export const handleCancel = ({ onClose }) => {
+export const handleCancel = ({ onClose, returnUrl }) => {
   if (onClose) {
     return onClose;
   }
+  let destination = useSelector(selectors.app.returnUrl);
+  console.log(returnUrl);
```
rayzhou-bit, line 57: console log should be removed

src/editors/containers/VideoEditor/index.jsx
```diff
@@ -31,6 +32,7 @@ export const VideoEditor = ({
       <EditorContainer
         getContent={fetchVideoContent()}
         onClose={onClose}
+        returnUrl={returnUrl}
```
rayzhou-bit, line 35: same question as above

rayzhou-bit (changes requested):
looks good, just some minor things to consider/change

src/editors/hooks.js
```diff
@@ -17,13 +17,18 @@ export const navigateTo = (destination) => {
...
   analyticsEvent,
   analytics,
 }) => () => {
   if (process.env.NODE_ENV !== 'development' && analyticsEvent && analytics) {
     sendTrackEvent(analyticsEvent, analytics);
   }
+  if (returnFunction) {
+    returnFunction();
+    return;
+  }
```
rayzhou-bit, line 31: codecov mentions you're missing tests here, but otherwise, it looks good to me!

rayzhou-bit (changes requested):

rayzhou-bit (approved):

PR #345, opened 2023-06-12
feat: parse solution text with preserve order

rayzhou-bit:
For the XMLParser, the `preserveOrder` option needs to be set to true for us to capture solutions properly when there are nested tags. Otherwise, the parser will bundle certain texts together in an undesired fashion.

https://2u-internal.atlassian.net/browse/TNL-10728

### DIFF
--- a/src/editors/containers/ProblemEditor/data/OLXParser.js
+++ b/src/editors/containers/ProblemEditor/data/OLXParser.js
@@ -63,17 +63,7 @@ export const stripNonTextTags = ({ input, tag }) => {
 export class OLXParser {
   constructor(olxString) {
     this.problem = {};
-    this.questionData = {};
-    const questionOptions = {
-      ignoreAttributes: false,
-      alwaysCreateTextNode: true,
-      numberParseOptions: {
-        leadingZeros: false,
-        hex: false,
-      },
-      preserveOrder: true,
-      processEntities: false,
-    };
+    this.problemArray = [];
     const parserOptions = {
       ignoreAttributes: false,
       alwaysCreateTextNode: true,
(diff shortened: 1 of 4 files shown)


src/editors/containers/ProblemEditor/data/OLXParser.js
```diff
@@ -63,25 +63,27 @@ export const stripNonTextTags = ({ input, tag }) => {
 export class OLXParser {
   constructor(olxString) {
     this.problem = {};
-    this.questionData = {};
-    const questionOptions = {
+    this.problemArray = [];
+    const parserOptions = {
```
connorhaugh, line 67: there should be a comment here explaining the three different options and why they are different

src/editors/containers/ProblemEditor/data/OLXParser.js
```diff
@@ -366,12 +367,12 @@ export class OLXParser {
         questionArray.push(tag);
       } else if (responseKeys.includes(tagName)) {
         /* <label> and <description> tags often are both valid olx as siblings or children of response type tags.
-         They, however, do belong in the question, so we append them to the question.
+          They, however, do belong in the question, so we append them to the question.
```
connorhaugh, line 370: You can remove this extra space

KristinAoki (changes requested):
A lot of your changes have already been made to `main`. I would rebase with `main` and check the functionality.

rayzhou-bit:
> A lot of your changes have already been made to `main`. I would rebase with `main` and check the functionality.

I tested the issue again and it is resolved in some ways. After a rebase attempt, the functionality that needs to be added is setting `trimValues` to false so the spaces before and after tags are kept in rich text fields.

rayzhou-bit:
I'm going to create a new PR off of main and close this one out. I must have been working in an older one that matched prod behavior at the time.

PR #346, opened 2023-06-12
feat: update returnToUnit to include api response

KristinAoki:
Related JIRA Ticket: [TNL-10796](https://2u-internal.atlassian.net/browse/TNL-10796)

This PR returns the xblock API response back to the parent component when `returnFunction` is defined. This is necessary when the editor data is changed and the change needs to be reflected on save.


kenclary (approved):
lgtm

PR #349, opened 2023-06-15
feat: add ErrorAlert to exports

KristinAoki:


connorhaugh (approved):
Does this mean we should port this component to pargagon?

KristinAoki:
The underlying element, `Alert`, is already part of paragon, so adding it would be redundant. The export is so that I don't have to duplicate the dismiss functionality because it is already built here.

PR #355, opened 2023-06-22
feat: more/correct v2 url handling. TNL-10742

kenclary:
feat: more/correct v2 url handling. TNL-10742

### DIFF
--- a/src/editors/data/services/cms/urls.js
+++ b/src/editors/data/services/cms/urls.js
@@ -7,26 +7,39 @@ export const unit = ({ studioEndpointUrl, unitUrl }) => (
 );
 
 export const returnUrl = ({ studioEndpointUrl, unitUrl, learningContextId }) => {
-  if (learningContextId && learningContextId.includes('library-v1')) {
+  if (learningContextId && learningContextId.startsWith('library-v1')) {
     // when the learning context is a v1 library, return to the library page
     return libraryV1({ studioEndpointUrl, learningContextId });
   }
+  if (learningContextId && learningContextId.startsWith('lib')) {
+    // when it's a v2 library, there will be no return url (instead a closed popup)
+    throw new Error('Return url not available (or needed) for V2 libraries');
+  }
   // when the learning context is a course, return to the unit page
-  return unitUrl ? unit({ studioEndpointUrl, unitUrl }) : '';
+  if (unitUrl) {
+    return unit({ studioEndpointUrl, unitUrl });
+  }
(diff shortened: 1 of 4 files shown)


src/editors/data/services/cms/urls.js
```diff
@@ -7,26 +7,34 @@ export const unit = ({ studioEndpointUrl, unitUrl }) => (
...
     ? `${studioEndpointUrl}/xblock/${blockId}`
-    : `${studioEndpointUrl}/api/xblock/v2/xblocks/${blockId}`
+    : `${studioEndpointUrl}/api/xblock/v2/xblocks/${blockId}/fields/`
 );
 
 export const blockAncestor = ({ studioEndpointUrl, blockId }) => (
-  `${block({ studioEndpointUrl, blockId })}?fields=ancestorInfo`
+  blockId.startsWith('block-v1')
+    ? `${block({ studioEndpointUrl, blockId })}?fields=ancestorInfo`
+    : '' // this url only need to get info to build the return url, which isn't used by V2 blocks
```
connorhaugh, line 31: Shouldn't we instead just not make this call in that case?

kenclary: <shrug> maybe, but it also makes sense to check the blockId here, until we refactor all this after moving everything to popups.

connorhaugh: specifically: wouldn't this lead to an error, as it would attempt to make a an axios call at ''?

src/editors/data/services/cms/urls.js
```diff
@@ -7,26 +7,34 @@ export const unit = ({ studioEndpointUrl, unitUrl }) => (
...
 
 export const returnUrl = ({ studioEndpointUrl, unitUrl, learningContextId }) => {
-  if (learningContextId && learningContextId.includes('library-v1')) {
+  if (learningContextId && learningContextId.startsWith('library-v1')) {
     // when the learning context is a v1 library, return to the library page
     return libraryV1({ studioEndpointUrl, learningContextId });
   }
+  if (learningContextId && learningContextId.startsWith('lib')) {
+    // when it's a v2 library, there will be no return url (instead a closed popup)
+    return '';
```
connorhaugh, line 16: Same with here?

src/editors/data/services/cms/urls.js
connorhaugh: or maybe we want this to error loudly if it is used, and instead of returning '' we should raise an error.

PR #356, opened 2023-06-24
feat: parser changes

rayzhou-bit:
This PR makes the following changes to the problem editor parser:
- `<div class="detailed-solution">` should not be required under the `<solution>` tag. The parser no longer requires this.
- The question field was only taking label, description, h1, h2, h3, h4, h5, h6, div, p and pre tags. This means tags like `<table>` and `` are not parsed and get removed when re-editing. These changes fixes this.
- These changes also fix the issue of new lines of `<p>Explanation</p>` popping up after saving and re-editing.

This PR was made due to several issues that came in where parts of the problem was missing or `<p>Explanation</p>` was added.
https://2u-internal.atlassian.net/browse/TNL-10832
https://2u-internal.atlassian.net/browse/TNL-10840
https://2u-internal.atlassian.net/browse/TNL-10833

### DIFF
--- a/src/editors/containers/ProblemEditor/components/EditProblemView/AnswerWidget/AnswersContainer.test.jsx
+++ b/src/editors/containers/ProblemEditor/components/EditProblemView/AnswerWidget/AnswersContainer.test.jsx
@@ -18,7 +18,7 @@ jest.mock('@edx/frontend-platform/i18n', () => ({
   getLocale: jest.fn(),
 }));
 
-jest.mock('./AnswerOption', () => function () {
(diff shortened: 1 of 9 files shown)


kenclary (approved):
yeah, looks good

jesperhodge (changes requested):
Hey, from the code itself, this looks good. Both as a reviewer and someone working on the code later, I would need to know what those bugs are and what the expected behavior is. So each of these fixed problems needs a test case, to ensure they don't reappear and that they are fixed in the first place.

jesperhodge (approved):

PR #326, opened 2023-05-04
[FAL-3383] Implement new video UX flow on new video editor

ChrisChV:
This branch adds all the navigation for the new video editor flow.

Internal Jira task: [FAL-3383](https://tasks.opencraft.com/browse/FAL-3383)

**Testing instructions**

- Use this branch https://github.com/openedx/edx-platform/pull/32024 and this branch https://github.com/openedx/frontend-app-course-authoring/pull/498
- Login on studio with a super user.
- We need to test the flow before to create any video.
- On a course create a unit and create a new video. Verify that it redirects to the video gallery, but after loading, it redirects to the video upload page.
- Go to [Django Admin](http://localhost:18010/admin/edxval/coursevideo/) and create a conection with a video and the Demo course.
    - Edit the duration of the videos.
- On a course create a unit and create a new video. Verify that it redirects to the video gallery.
- Check the cancel button
- Create new video and on the video gallery, select a video and go to the next page.
- Verify that the Video ID and de duration are filled.
- Click on the `Replace video button`.

To test the video upload, you need to configure an S3, but there is an easy way to test this

- Read [this comment](https://github.com/openedx/frontend-lib-content-components/pull/326/files#diff-8c87d7bec267a780f66f8210f786b97cec0d8bdafacc162a220d13343393c286R30)
- Update the [videos_handler](https://github.com/openedx/edx-platform/blob/d622db59b7bfe2e4c11fc11b7cb2d475cfda551b/cms/djangoapps/contentstore/views/videos.py#L188) function on edx-paltform to return overwritten data. You can use this [gist as reference](https://gist.github.com/ChrisChV/e662a5bdb4145a80606295f5b677d5b5)
- Create a video and on the video gallery, click on the upload button.
- Upload any video
- Verify that the Video ID and de duration are filled.

**Author notes**

- The video upload only works with an S3 configuration.
- I added a temporal line to test the video upload without an S3 configuration: [Comment](https://github.com/openedx/frontend-lib-content-components/pull/326/files#diff-8c87d7bec267a780f66f8210f786b97cec0d8bdafacc162a220d13343393c286R30)
- This branch was internal reviewed here: https://github.com/open-craft/frontend-lib-content-components/pull/8#issuecomment-1523827421

**Release notes**

**IMPORTANT**: Delete [this line](https://github.com/openedx/frontend-lib-content-components/pull/326/files#diff-8c87d7bec267a780f66f8210f786b97cec0d8bdafacc162a220d13343393c286R30) before merge

PR #366, opened 2023-08-01
Re-write the dropzone component and fix styling issues

farhaanbukhsh:
## Description

We are planning to use the Dropzone Paragone component here and also fix the style for URL video uploader. We are trying to keep the paragon component intact for it to be used.

Introduced new thumbnail

image

Introduced URL uploader field

image

We remove the custom Dropzone that was written initially and have used the paragon inbuilt one.

## Testing instructions

1. Set up the master devstack, make sure you have course-authoring-mfe(frontend-app-course-authoring) running
2. We have to clone this repo in {EDX_DEVSTACK}/src directory.
3. Change the branch being used to `open-craft:farhaan/fix-drag-drop-component` and run `make build`
4. Follow the steps mentioned in https://github.com/openedx/frontend-lib-content-components/pull/262 to set up video editor, make sure to `Step 3 - Step 7`
5. We need to upload a small video and connect it to the demo course as mentioned in https://github.com/openedx/frontend-lib-content-components/pull/326
6. We need to make changes to the edx-platform to dummy S3 connection as mentioned in the above PR.
7. Now add a video xblock in any unit, you will be redirected to the the course-authoring-mfe's video setting page.
8. Check the thumbnail here -- new addition
8. Go to "Replace Video"
9. Try to upload a video, you should see a spinner here.
10. You will see the new URL bar here as well -- this works like a search field.

## Other information

Private ref: https://tasks.opencraft.com/browse/BB-7524

Reviewer:

- [ ] @0x29a


src/editors/containers/VideoUploadEditor/index.jsx
```diff
@@ -11,77 +12,69 @@ import * as hooks from './hooks';
...
-  const { textInputValue, setTextInputValue } = hooks.uploader();
-  const onURLUpload = hooks.onVideoUpload();
+const URLUploader = () => (
+  <div className="d-flex flex-column flex-wrap">
+    <div className="justify-content-center align-self-center bg-light rounded-circle p-4">
+      <Icon src={FileUpload} className="text-muted" />
+    </div>
+    <div className="d-flex align-self-center justify-content-center flex-wrap flex-column pt-5">
+      <span style={{ fontSize: '20px' }}><FormattedMessage {...messages.dropVideoFileHere} /></span>
+      <span className="align-self-center" style={{ fontSize: '12px' }}><FormattedMessage {...messages.info} /></span>
```
xitij2000, line 22: I think using FormattedMessage like this doesn't provide any real benefit over just using `intl.formatMessage`, so I'd recommend that instead.

src/editors/containers/VideoUploadEditor/index.jsx
```diff
@@ -11,77 +12,69 @@ import * as hooks from './hooks';
...
-    multiple: false,
-    onDrop: (acceptedFiles) => {
-      if (acceptedFiles.length > 0) {
-        const uploadfile = acceptedFiles[0];
-        onUpload(uploadfile);
-      }
-    },
-  });
+export const VideoUploader = ({ onUpload }) => {
+  const { textInputValue, settextInputValue } = hooks.uploader();
```
xitij2000, line 29:
```suggestion
  const { textInputValue, setTextInputValue } = hooks.uploader();
```

xitij2000: I am not seeing any benefit in calling hooks.uploader, which then calls testInputValue when then calls useState. 

Just directly have React.useState here.

src/editors/containers/VideoUploadEditor/index.jsx
```diff
@@ -120,7 +107,7 @@ const VideoUploadEditor = (
               onClick={handleCancel}
             />
           </div>
-          <VideoUploader onUpload={handleDrop} errorMessage={errorMessage} intl={intl} />
+          <VideoUploader onUpload={handleDrop} intl={intl} />
```
xitij2000, line 110: Why are we passing `intl` here? We can just use `useIntl` in that component.

src/editors/containers/VideoUploadEditor/index.jsx
```diff
@@ -11,77 +12,69 @@ import * as hooks from './hooks';
...
-    onDrop: (acceptedFiles) => {
-      if (acceptedFiles.length > 0) {
-        const uploadfile = acceptedFiles[0];
-        onUpload(uploadfile);
-      }
-    },
-  });
+export const VideoUploader = ({ onUpload }) => {
+  const { textInputValue, settextInputValue } = hooks.uploader();
+  const onURLUpload = hooks.onVideoUpload();
```
xitij2000, line 30: This is about onVideoUpload. 

That function is doing things it shouldn't. 

1. Instead of importing store directly you can use the `useStore` hook. 
2. If you want to pass the store as a function argument, just call that function with `useSelector` i.e. `useSelector(postUploadRedirect`. 

I think what you should do instead is, call the selector `postUploadRedirectUrl` as follows:

```
const postUploadRedirectUrl = createSelector([learningContextId, blockId], (contextId, blockId) =>`/course/${learningContextId}/editor/video/${blockId}?selectedVideoUrl=${videoUrl}`). 
```
Then use it as:
```
const redirectUrl = useSelector(postUploadRedirectUrl);
...
// In some component:
//onClick={() => navigateTo(redirectUrl)
```

src/editors/containers/VideoUploadEditor/index.jsx
```diff
@@ -11,77 +12,69 @@ import * as hooks from './hooks';
...
+  const URLUploaderComponent = URLUploader();
 
-  const handleInputChange = (event) => {
-    setTextInputValue(event.target.value);
+  const handleInputChange = (value) => {
```
xitij2000, line 33: This function isn't doing anything, just set onChange to setTextInputChange directly.

src/editors/containers/VideoUploadEditor/index.jsx
```diff
@@ -11,77 +12,69 @@ import * as hooks from './hooks';
...
-    );
-  }
+  const handleProcessUpload = async ({
+    // eslint-disable-next-line no-unused-vars
+    fileData, requestConfig, handleError,
+  }) => {
+    try {
+      await onUpload(fileData);
+    } catch (error) {
+      console.log(error);
```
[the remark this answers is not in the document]
xitij2000: Currently the errors are being swallowed. I think there should be some response to users.

src/editors/containers/VideoUploadEditor/index.jsx
```diff
@@ -11,77 +12,69 @@ import * as hooks from './hooks';
...
   };
 
-  if (errorMessage) {
-    return (
-      <div className="d-flex flex-column justify-content-center align-items-center text-center error-message">{errorMessage}</div>
-    );
-  }
+  const handleProcessUpload = async ({
+    // eslint-disable-next-line no-unused-vars
+    fileData, requestConfig, handleError,
```
xitij2000, line 43: If the params are not used just remove them?

src/editors/containers/VideoUploadEditor/index.jsx
```diff
@@ -11,77 +12,69 @@ import * as hooks from './hooks';
...
-          </button>
-        </div>
+      <Dropzone
+        accept={{ 'video/*': ['.mp4', '.mov'] }}
+        onProcessUpload={handleProcessUpload}
+        inputComponent={URLUploaderComponent}
+      />
+      <div className="d-flex video-id-prompt">
+        <SearchField
+          placeholder="Paste your video ID or URL"
```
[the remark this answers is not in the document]
xitij2000: Additionally, why is this a search field?

src/editors/containers/VideoUploadEditor/index.jsx
```diff
@@ -95,17 +88,11 @@ const VideoUploadEditor = (
   const {
     loading,
     setLoading,
-    errorMessage,
-    setErrorMessage,
   } = hooks.uploadEditor();
   const handleCancel = editorHooks.handleCancel({ onClose });
```
xitij2000, line 92: It seems if you provide onClose as an argument to this function it returns that function, so it is doing nothing here. This function take onClose and returns onClose. So this call can be skipped.

src/editors/containers/VideoUploadEditor/index.jsx
```diff
@@ -95,17 +88,11 @@ const VideoUploadEditor = (
   const {
     loading,
     setLoading,
-    errorMessage,
-    setErrorMessage,
   } = hooks.uploadEditor();
```
xitij2000, line 91: This can also be simplified to a React.useState call.

src/editors/containers/VideoUploadEditor/hooks.js
```diff
@@ -54,22 +49,17 @@ export const onVideoUpload = () => {
...
-  } else {
-    const errorMsg = 'Video must be an MP4 or MOV file';
-    setErrorMessage(errorMsg);
-  }
+  uploadVideo({
+    supportedFiles: [newFile],
+    setLoadSpinner: setLoading,
+    postUploadRedirect: onVideoUpload(),
+  });
 };
```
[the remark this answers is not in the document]
xitij2000: @0x29a I think the dropzone component already does validation, so it might be OK to skip it here. 

That said, I don't understand what this function is doing. Are browsers not filling in the mime type? I tried a couple of files and the type is already set correctly, so you can just use the original file instead of creating a new one. I agree that it's no longer a validator, so the name should change. I think removing that function can calling uploadVideo directly would be better.

xitij2000 (changes requested):

src/editors/containers/VideoUploadEditor/index.scss
```diff
@@ -4,11 +4,15 @@
...
   }
+}
 
-  .file-upload {
-    width: 56px;
-    height: 56px;
-  }
+.dropzone-file-container {
+  height: 100vh;
+}
```
xitij2000, line 11: I don't see this being used.

src/editors/containers/VideoUploadEditor/index.scss
```diff
@@ -32,12 +34,22 @@
...
     border: none !important;
     background-color: #FFFFFF;
   }
+
+  .prompt-button {
+    background: rgba(239, 234, 247, 0.70);
+  }
 }
 
 .error-message {
```
xitij2000, line 55: A lot of this file contains magic values. Can you instead use variables so that it picks up the correct theme?

You can see the variables here: https://paragon-openedx.netlify.app/foundations/css-utilities/
Find the matching variable for what you want. I think you might need to look at the edX theme, since I think figma designs generally use the edX theme.

xitij2000 (changes requested):

src/editors/containers/VideoUploadEditor/hooks.js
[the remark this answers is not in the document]
farhaanbukhsh: @0x29a @xitij2000  you are write we no longer need this function since Dropzone can handle this.

src/editors/containers/VideoUploadEditor/index.jsx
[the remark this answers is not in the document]
farhaanbukhsh: Probably but we are not making it reach the user here. May be in the next iteration.

src/editors/containers/VideoUploadEditor/index.jsx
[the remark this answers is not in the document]
farhaanbukhsh: I have changed this to a inputGroup now

src/editors/containers/VideoUploadEditor/index.scss
farhaanbukhsh: I have tried removing as much as I can :)

src/editors/containers/VideoUploadEditor/index.jsx
farhaanbukhsh: It is going with the editor component which this component is wrapped in, FWIW it's telling the parent component about it. I am keeping this for now.

src/editors/containers/VideoUploadEditor/index.jsx
farhaanbukhsh: useIntl is not available to us as pointed out in https://github.com/openedx/frontend-lib-content-components/blob/main/docs/decisions/0005-internal-editor-testability-decisions.md#additional-considrations

xitij2000: I think it's available now, I've used it in my latest PR.

src/editors/containers/VideoUploadEditor/index.jsx
xitij2000: I'm not sure I understand what you mean? `handleCancel` is just a regular function. I think you can replace this with
`  const handleCancel = onClose;` and it will work the same. 

It's not part of this PR, so it's OK to leave it for now.

src/editors/containers/VideoUploadEditor/index.jsx
```diff
@@ -11,77 +12,73 @@ import * as hooks from './hooks';
 import messages from './messages';
 import * as editorHooks from '../EditorContainer/hooks';
 
-export const VideoUploader = ({ onUpload, errorMessage }) => {
-  const { textInputValue, setTextInputValue } = hooks.uploader();
-  const onURLUpload = hooks.onVideoUpload();
+const URLUploader = ({ intl }) => (
```
xitij2000, line 15: You can use `const intl = useIntl()` here. I think the advice against that is outdated since this uses a newer version of platform now which includes that.

src/editors/containers/VideoUploadEditor/hooks.js
```diff
@@ -52,30 +18,21 @@ export const onVideoUpload = () => {
...
-    uploadVideo({
-      supportedFiles: [newFile],
-      setLoadSpinner: setLoading,
-      postUploadRedirect: onVideoUpload(),
-    });
-  } else {
-    const errorMsg = 'Video must be an MP4 or MOV file';
-    setErrorMessage(errorMsg);
-  }
+export const useUploadVideo = async ({
```
xitij2000, line 21: This is not a react hook, and so should not be named starting with use. 
Additionally, its not adding much so perhaps the dispatch call can be moved directly into the component?
Is the idea to wrap it here to make it easier to mock?

xitij2000 (changes requested):

src/editors/containers/VideoUploadEditor/index.jsx
farhaanbukhsh: This is required to get the close button working on the editor since this is passed as a parameter to a higher-order function.

src/editors/data/images/videoThumbnail.svg
xitij2000, line 1: I think the right place to add this icon will be in paragon. 

It is a pretty simple process that you can see [here](https://github.com/openedx/paragon/tree/master/icons#readme). You'll just need to put this svg in the correct folder, run build in the icon folder and make a PR based on the changes. 

I think you can remove this icon for now and put that in a separate PR?

src/editors/containers/VideoUploadEditor/index.scss
```diff
@@ -4,45 +4,45 @@
...
-  border: 1px solid #707070;
-  width: 308px;
-
-  input {
-    border: none !important;
-    width: 90%;
-  }
 
   input::placeholder {
     color: #454545;
```
xitij2000, line 31: This corresponds with $gray-700, you can use that here.

src/editors/containers/VideoUploadEditor/index.scss
```diff
@@ -4,45 +4,45 @@
...
+    // color: #5E35B1;
+    font-family: 'Inter';
+    font-weight: 500;
+    word-wrap: break-word;
```
xitij2000, line 35: I don't think you need to specify font-family here. Inter is the font used by the [edx theme ](https://github.com/edx/brand-edx.org/blob/d4e4fa4dcd06b2fec01af97121ebd278ec95c89b/paragon/_variables.scss#L419), it is not available in the openedx theme. So if you use the edx branding variant, this font will probably already be applied, but for the the standard branding the font will be different.

xitij2000 (approved):
:+1:

- [x] I tested this: tested on devstack
- [x] I read through the code

kenclary (approved):
ok, looks good

PR #369, opened 2023-08-08
fix: error return while editor loading

KristinAoki:
1. Go to link in studio
2. Click “edit” next to text component.
3. expect the page to load


rayzhou-bit (approved):
lgtm

PR #372, opened 2023-08-09
fix: blockId checks for v1 or v2 libraries

KristinAoki:
1. open up a library block
2. should render as expected
3. test cancel and save capabilities


rayzhou-bit (approved):
lgtm

PR #374, opened 2023-08-11
fix: selectedVideo undefined error

KristinAoki:
Fix for [new relic error](https://one.newrelic.com/alerts-ai/issue?account=88178&duration=259200000&state=a8934090-9fec-dbdf-14e5-7b0f28b91a5a) type error

Test
1. Open a video block
2. replace the video
3. block should load as expected with video id
4. Navigate back to unit page
5. create new video bloack
6. select video
7. block should load as expected with video id


rayzhou-bit (approved):
tested this on local devstack and it looks good!

PR #382, opened 2023-08-28
docs: document jest troubleshooting

jesperhodge:
This is a documentation addition, since the last time we included this library in a new project, the jest pipeline failed. This documents what is necessary to fix that.

https://2u-internal.atlassian.net/browse/TNL-10927


kenclary (approved):
I think this covers it.

PR #387, opened 2023-09-11
fix: sort items in typeahead dropdown

KristinAoki:
JIRA Ticket: [TNL-11018](https://2u-internal.atlassian.net/browse/TNL-11018)

This PR sorts the items for the typeahead dropdown before the items are rendered into buttons.

### DIFF
--- a/src/editors/sharedComponents/TypeaheadDropdown/index.jsx
+++ b/src/editors/sharedComponents/TypeaheadDropdown/index.jsx
@@ -45,7 +45,7 @@ class TypeaheadDropdown extends React.Component {
       options = options.filter((option) => (option.toLowerCase().includes(strToFind.toLowerCase())));
     }
 
-    return options.map((opt) => {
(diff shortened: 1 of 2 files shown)


src/editors/sharedComponents/TypeaheadDropdown/index.jsx
```diff
@@ -220,6 +220,7 @@ class TypeaheadDropdown extends React.Component {
             data-testid="dropdown-container"
             className="dropdown-container mt-2 rounded bg-light-100 box-shadow-centered-1 mr-2"
             style={{ maxHeight: '300px', overflowY: 'scroll' }}
+            role="listbox"
```
connorhaugh, line 223: nit: i think you probably added this to add test, but seems no longer needed? Unless it is good for a11y?

connorhaugh (approved):
yep. Just a nit.

PR #392, opened 2023-09-14
fix: v2 libraries default open to advanced editor

KristinAoki:
JIRA Ticket: [TNL-11023](https://2u-internal.atlassian.net/browse/TNL-11023)

Currently all problems in v2 libraries open in the advanced editor. This is because the problem settings are being added to the wrapping `<problem>` tag and threw a "Misc. Attributes" error. This data is duplicated in the problem metadata. This PR adds a check to the OLX parser to only throw "Misc. Attributes" error if the attributes are not settings.

Testing

**Repeat test in a v2 library, v1 library, and course**

1. Create a new common problem, fill out fields, and save.
2. Edit the block and it should open in the simple editor.
4. Choose to open in the advanced editor.
5. Add an attribute to the `<problem>` tag (e.g guess="no") and save.
6. Edit the block and it should open in the advanced editor.
7. Create a new advanced problem and save.
8. Edit the block.
9. Should open in the advanced editor.

### DIFF
--- a/src/editors/containers/ProblemEditor/components/EditProblemView/SettingsWidget/settingsComponents/ResetCard.jsx
+++ b/src/editors/containers/ProblemEditor/components/EditProblemView/SettingsWidget/settingsComponents/ResetCard.jsx
@@ -14,6 +14,7 @@ export const ResetCard = ({
   // inject
   intl,
 }) => {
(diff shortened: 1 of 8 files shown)


kenclary (approved):
oo, good find

PR #393, opened 2023-09-14
fix: block url conditional

KristinAoki:
JIRA Ticket: [TNL-10971](https://2u-internal.atlassian.net/browse/TNL-10971)

This PR ensures that if the block id is for a v2 library it will us ethe xblock url for v2 libraries. All other block ids will use the generic xblock url.

Test

1. Open an existing v2 library block.
2. Should render without errors.
3. Open an existing course block.
4. Should render without errors.
5. Open an existing v1 library block.
6. Should render without errors

### DIFF
--- a/src/editors/data/services/cms/urls.js
+++ b/src/editors/data/services/cms/urls.js
@@ -29,9 +29,9 @@ export const returnUrl = ({
 };
 
 export const block = ({ studioEndpointUrl, blockId }) => (
-  blockId.includes('block-v1')
-    ? `${studioEndpointUrl}/xblock/${blockId}`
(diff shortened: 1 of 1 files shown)


connorhaugh:
In my head, this is an if else statement that we are simply inverting. Why is this change needed?

KristinAoki:
> In my head, this is an if else statement that we are simply inverting. Why is this change needed?

To solve the case listed in the ticket description
> This change fixed an issue with v1 libraries not loading: fix: blockId checks for v1 or v2 libraries (Pull Request #372 ) However, it does not consider the situation where library id contains ‘block-v1’.

connorhaugh (approved):
AH, a roundabout way of doing that.

PR #411, opened 2023-10-17
feat: initial editor layout + ui

connorhaugh:


src/editors/containers/EditorContainer/index.jsx
```diff
@@ -64,7 +64,7 @@ export const EditorContainer = ({
           />
         </div>
       </ModalDialog.Header>
-      <ModalDialog.Body className="pb-6">
+      <ModalDialog.Body className="pb-6 min-vh-100">
```
connorhaugh, line 67: Why was this added? I feel like it might be an issue for full screen modal stuff?

src/editors/data/services/cms/api.js
```diff
@@ -180,6 +180,29 @@ export const apiMethods = {
           license: module.processLicense(content.licenseType, content.licenseDetails),
         },
       };
+    } else if (blockType ==='library_content') {
+      response = {
+        category: blockType,
+        courseKey: learningContextId,
+        display_name: title, //"course-v1:edx+123+test"
+        // has_children: true,
```
connorhaugh, line 188: Why is this field commented out? presuming you are working on this atm

src/editors/data/services/cms/api.js
```diff
@@ -180,6 +180,29 @@ export const apiMethods = {
...
+        highlights_enabled_for_messaging: false,
+        highlights_preview_only: true,
+        id: blockId, //"block-v1:edx+123+test+type@library_content+block@880758e62c2542d5b45c944360dfa799"
+        metadata: {
+          allow_resetting_children: content.showReset,
+          candidates: content.candidates,
+          capa_type: 'any',
+          count: content.count,
+          //manual
+          //shuffle
```
connorhaugh, line 201: We'll want to uncomment these.
```suggestion
          manual
          shuffle
```

connorhaugh (commented):
Just some early feedback (I realize this is not fully ready for review yet.)

src/editors/containers/EditorContainer/index.jsx
rayzhou-bit: Without this, the dropdown can get cut off because the "body" of the editor doesn't extend all the way down

src/editors/containers/LibraryContentEditor/data/urls.js
```diff
@@ -0,0 +1,19 @@
+export const v1Libraries = ({ studioEndpointUrl }) => (
+  `${studioEndpointUrl}/api/contentstore/v1/home`
+);
+
+export const v2Libraries = ({ studioEndpointUrl }) => (
+  `${studioEndpointUrl}/api/libraries/v2/`
+);
+
+export const libraryMetadata = ({ studioEndpointUrl, libraryId }) => (
```
connorhaugh, line 9: note: v1 library metadata will come from a different API.

src/editors/containers/LibraryContentEditor/utils.js
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * gets the name of a library whether it is v1 or v2
+ * @param {[object]} library - library
+ * @returns {[string]} library name. returns an empty string if nothing is found
+ */
+export const getLibraryName = (library) => {
+  if (!library) {
+    return '';
+  }
+  return library.title || library.display_name || '';
```
connorhaugh, line 25: will object has no attribute errors be a problem here?

connorhaugh (commented):
What does the lifecycle of the library api methods look like? Do we only initialize them for library editors (I hope that is a yes). Just couldn't see that cleanly in the code.

src/editors/containers/LibraryContentEditor/data/urls.js
rayzhou-bit: I'll rename these urls and request apis to be specific to library version

src/editors/containers/LibraryContentEditor/BlocksSelector.jsx
```diff
@@ -0,0 +1,142 @@
...
+  );
+};
+
+RowCheckbox.defaultProps = {
+  row: {},
+};
+
+RowCheckbox.propTypes = {
+  row: PropTypes.shape({
+    getToggleRowSelectedProps: PropTypes.func.isRequired,
```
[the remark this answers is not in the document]
rayzhou-bit: `row.getToggleRowSelectedProps` is...The linter wasn't happy that I didn't mention it.

src/editors/containers/LibraryContentEditor/utils.js
rayzhou-bit: If there's no attribute, it should return `undefined` and go to the next item. This should essentially be
```
if (library.title) return library.title;
if (library.display_name) return library.display_name;
return '';
```

rayzhou-bit:
> What does the lifecycle of the library api methods look like? Do we only initialize them for library editors (I hope that is a yes). Just couldn't see that cleanly in the code.

You should be able to see these in the `useEffect` calls in the hooks.js file.
1. When editor is initialized, lists of v1 and v2 libraries are fetched.
2. When a library is selected (including when opening an editor with an already selected library), the blocks of the library is fetched.

rayzhou-bit:
Oof I left the description half finished. I'll update that in a bit and I'll make sure to add that screenshot.

src/editors/data/constants/requests.js
```diff
@@ -29,4 +29,9 @@ export const RequestKeys = StrictDict({
   uploadImage: 'uploadImage',
   fetchAdvancedSettings: 'fetchAdvancedSettings',
   fetchVideoFeatures: 'fetchVideoFeatures',
+  fetchV1Libraries: 'fetchV1Libraries',
+  fetchV2Libraries: 'fetchV2Libraries',
+  fetchV2LibraryMetadata: 'fetchV2LibraryMetadata',
+  fetchV2LibraryContent: 'fetchV2LibraryContent',
+  fetchBlockContent: 'fetchBlockContent',
```
connorhaugh, line 36: It looks like this api call is not made or used atm, right?

connorhaugh:
Hey -- it looks great but a couple of bugs from manual testing:

1. State preservation
-   Create A V2 lib
-   Associate it with a lib content block
-  Set it to "Show the selected blocks of the library to all users." and select 1 block
- Save & exit
- **Expected behavior** Returning to the editor shows "Show the selected blocks of the library to all users." and the block is selected
- **Actual behavior** Returning to the editor shows "Show a number of components to display at random from the library." 

2. Saving Errors

- Create a v1 lib
- Associate it with a library content block.
- set the max_count  to be anything
- Save
- **Expected behavior** Save is successful and returns you to the unit page.
- **Actual behavior**  you get a full page error saying Cannot read properties of undefined (reading 'toString')

3. Follow on bug (not a blocker for you or dealt with in this PR just have to confirm it works when we merge) candidates aren't interpreted in the "view" page.

src/editors/data/constants/requests.js
rayzhou-bit: `fetchV2LibraryContent` is used but `fetchV2LibraryMetadata` is not. This was originally for fetching v1 library metadata (before I renamed some things to v2). V1 library metadata is needed to get v1 library version. I'll need to put this back in.

src/editors/containers/LibraryContentEditor/index.test.jsx
```diff
@@ -0,0 +1,62 @@
...
+    intl: {
+      formatMessage: (content) => content.defaultMessage,
+    },
+  };
+
+  it('Renders a spinner when loading', () => {
+    const { getByTestId } = renderComponent({ ...props, blockFinished: false });
+    expect(getByTestId('librarycontenteditor-loadingspinner')).toBeTruthy();
+  });
+  it('Renders as expected once loaded', () => {
```
jesperhodge, line 55: Also on this component level, we need to consider what should happen if the loading has finished but the block fails. What should be rendered? Should child components be rendered if there is no block to render? Should we throw an error that NewRelic will pick up? - If we do still render child components, will child components throw unwanted errors?

I would include a test for it with the desired behavior, and probably the desired behavior warrants some thinking about what should happen.

- That's my first-glance take, maybe you already dealt with everything in the hook. It wouldn't hurt having an additional test here anyway, so we can see if a future change inside the component breaks something.

src/editors/containers/LibraryContentEditor/hooks.test.js
```diff
@@ -0,0 +1,279 @@
...
+        title: 'v2libtitle',
+      },
+    ];
+    const v1Libraries = [
+      {
+        display_name: 'myv1library',
+        library_key: 'v1id',
+      },
+    ];
+    it('should fetch v2 and v1 libraries when block finishes loading', () => {
```
jesperhodge, line 74: A blank line between `it` statements is somehow just essential for me. At some point I just start seeing a wall of text otherwise.

src/editors/containers/LibraryContentEditor/data/requests.js
```diff
@@ -0,0 +1,97 @@
+import { StrictDict } from '../../../utils';
```
jesperhodge, line 1: A lot here seems not covered by tests. Probably warrants a `requests.test.js` file.

src/editors/containers/LibraryContentEditor/index.test.jsx
connorhaugh: I think we want whatever is the most simple here in terms of behavior @rayzhou-bit , but +1 to having a unit test here.

src/editors/data/constants/requests.js
connorhaugh: We acttually need none of:

fetchV2LibraryContent 
fetchV2LibraryContent

as we shouldn't be getting the content from the library.

rayzhou-bit: `fetchV2LibraryContent` points to the `/api/libraries/v2/${libraryId}/blocks/` endpoint. We want this for library blocks right? Do I just need to rename this to `fetchV2LibraryBlocks` or is there something else I'm missing?

rayzhou-bit:
> 1. State preservation
>Expected behavior Returning to the editor shows "Show the selected blocks of the library to all users." and the block is selected

I'm not getting a `manual` in the api call.

src/editors/data/constants/requests.js
connorhaugh: I'm realizing there is more to this than I may have let on. from a UI perspective.
There are three-ish cases we need to consider here:

1. A user clicks edit on a library block which already has a source library_id then either:
- they view their selected blocks in the library.
- they click on a new library.
3. A user clicks  edit on a library block which does not already have a souce library_id.

In theory, in the first case, we need to show the library content block's children (not the library's) to be available for selection.

the last two cases are a conundrum. We know that child blocks of a library and child blocks of a library content block have different ids. Therefore, it is fruitless to attempt to do block selection in those cases. How can we fix this? I think we should do this as a fast-follow, but definitely this is a blocker to rollout.

rayzhou-bit: Just checking to see if I understand this correctly... Is the library content block's children the candidates of what might be an older version of the selected library?

kdmccormick:
We will be implementing static reference without a wrapper block. For example, instead of a course library_content block wrapping a library problem block, we will have a course problem block with a [pointer to the upstream_block](https://github.com/openedx/edx-platform/pull/34322). Randomized content will still need a wrapper block, which may be the library_content block, or may be a new block.

Thanks for all your work on this, folks. I'll pass this PR along to the team working on the new library-reference frontends so that they have the opportunity to reuse some of this work.

PR #428, opened 2023-12-08
fix: misssing part of paid plugin configuration

kenclary:
fix: misssing part of paid plugin configuration

### DIFF
--- a/src/editors/sharedComponents/TinyMceWidget/pluginConfig.js
+++ b/src/editors/sharedComponents/TinyMceWidget/pluginConfig.js
@@ -31,6 +31,8 @@ const pluginConfig = ({ isLibrary, placeholder, editorType }) => {
         image,
         imageTools,
         quickToolbar,
+        plugins.a11ychecker,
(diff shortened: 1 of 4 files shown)


src/editors/sharedComponents/TinyMceWidget/pluginConfig.js
KristinAoki, line 1: add `buttons.a11yChecker` to quickToolbar list

KristinAoki (changes requested):

src/editors/sharedComponents/TinyMceWidget/pluginConfig.js
kenclary: which of the regular toolbar, the quick insert toolbar, and the quick selection toolbar should it be in?

do we have any guidance about where in the toolbar(s) it should be?

kenclary: answered by Product (in slack)

KristinAoki (approved):

PR #436, opened 2023-12-14
chore: disable tinymce plugins

jesperhodge:
We have observed errors in the Editor when developing locally when the plugins are enabled. They do not occur on stage or prod for us, but we have a theory that they will occur when the openedx stubbed version of the plugins is used. So we will disable for now and work on a fix.


bszabo (approved):
lgtm

PR #437, opened 2023-12-14
Fix  disable tinymce powerpaste and a11ychecker plugins

jesperhodge:
Previous commit didn't create a release. So I prefixed the commit with "fix: ".


bszabo (approved):
trivial change in a comment only - fine

PR #444, opened 2023-12-19
fix  editor deleting description

jesperhodge:
internal issue: https://2u-internal.atlassian.net/browse/TNL-11311

This fixes a bug where the editor was deleting the <description> OLX tag when editing in the simple editor and then saving.
Also the description was being converted to em for the simple editor, but then not converted back. However, the xblock renders label and then em in reverse order for some reason.
To fix it, the em gets converted back to description now, but not for every em tag (added a class "olx_description" for the tags that should be converted).

Testing:
- Copy the OLX below to a new advanced blank problem and save.
- The description should be visible in the course outline.
- Open problem again in simple editor. The content of the description tag should be visible as italics.
- save and check course outline. The content of the description tag should be visible and in the right order.
- open problem again. convert to advanced problem. The description tag should be back.
- Test a few other things concerning converting OLX that you expect to be working with advanced and simple editor. We want to do our best to make sure that we don't break anything else with this.

```
<problem>
  <choiceresponse>
    <label>Learning about the benefits of preventative health care can be
    particularly difficult.</label>
    <description>Check all of the options below that might be reasons why.</description>
    <checkboxgroup>
      <choice correct="true">A large amount of time passes between
       undertaking a preventative measure and seeing the result.</choice>
      <choice correct="false">Non-immunized people will always fall sick.</choice>
      <choice correct="true">If others are immunized, fewer people will fall
       sick regardless of a particular individual's choice to get immunized
       or not.</choice>
      <choice correct="true">Trust in health care professionals and
       government officials is fragile.</choice>
    </checkboxgroup>
    <solution>
      <div class="detailed-solution">
        <p>Explanation</p>
        <p>People who are not immunized against a disease might still not
         fall sick from the disease. If someone is trying to learn whether
         or not preventative measures against the disease have any impact,
         he or she might see these people and conclude, since they have
         remained healthy despite not being immunized, that immunizations
         have no effect. Consequently, he or she would tend to believe that
         immunization (or other preventative measures) have fewer benefits
         than they actually do.</p>
      </div>
    </solution>
  </choiceresponse>
</problem>
```

### DIFF
--- a/src/editors/containers/ProblemEditor/data/OLXParser.js
+++ b/src/editors/containers/ProblemEditor/data/OLXParser.js
@@ -487,7 +487,12 @@ export class OLXParser {
       }
     });
     const questionString = this.richTextBuilder.build(questionArray);
-    return questionString.replace(/<description>/gm, '<em>').replace(/<\/description>/gm, '</em>');
+    const res = this.replaceOlxDescriptionTag(questionString);
+    return res;
+  }
+
+  replaceOlxDescriptionTag(questionString) {
+    return questionString.replace(/<description>/gm, '<em class="olx_description">').replace(/<\/description>/gm, '</em>');
   }
 
   /** getHints()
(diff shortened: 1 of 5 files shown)


connorhaugh (approved):
Local testing didn't raise any issues. LGTM

PR #445, opened 2023-12-20
refactor: abstract XML parser output operations

jesperhodge:
- fix: question parser
- fix: add class olx_description and parse html with this class back to description tag
- refactor: abstract out operations on XML parser output and add fine-grained tests


kenclary (approved):
looks great

PR #453, opened 2024-01-09
fix: url for video uploads

KristinAoki:
JIRA Ticket: [TNL-11269](https://2u-internal.atlassian.net/browse/TNL-11269)

> Upload a new video to the block. Note that there is a video id put into the video URL field, which is wrong.

The query parameter that controls whether the id in the url is a video Id or url was always set to be a url even when a video was uploaded instead of a url. Now the query parameter is set dynamically so the appropriate value is add to the url.

Testing
1. Add a video block
2. Click "Upload or embed new video"
3. Add a video url and submit
4. Url should load in url field
5. Click "Replace video"
6. Click "Upload or embed new video"
7. Use dropzone to upload new video file
8. The new video's id should be in the Video Id field.

### DIFF
--- a/src/editors/containers/VideoUploadEditor/hooks.js
+++ b/src/editors/containers/VideoUploadEditor/hooks.js
@@ -7,15 +7,15 @@ export const {
   navigateTo,
 } = appHooks;
 
-export const postUploadRedirect = (storeState) => {
+export const postUploadRedirect = (storeState, uploadType) => {
   const learningContextId = selectors.app.learningContextId(storeState);
   const blockId = selectors.app.blockId(storeState);
-  return (videoUrl) => navigateTo(`/course/${learningContextId}/editor/video/${blockId}?selectedVideoUrl=${videoUrl}`);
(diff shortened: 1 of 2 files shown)


src/editors/containers/VideoUploadEditor/hooks.js
```diff
@@ -7,15 +7,15 @@ export const {
   navigateTo,
 } = appHooks;
 
-export const postUploadRedirect = (storeState) => {
+export const postUploadRedirect = (storeState, uploadType) => {
```
jesperhodge, line 10: Can you add a default value here? e.g. `uploadType = selectedVideoUrl`?

since `${undefined}` evaluates to "undefined" and we get a 404 if someone doesn't add the parameter.

Unless you want a 404 produced in this case so we know something's up.

jesperhodge (changes requested):
I followed your testing instructions. In the last step when I tried uploading a video via dropzone, I got a not-ending loading spinner and in the console a 500 error from http://localhost:18010/course/course-v1:edx+videostestaz+2023/videos/. It's maybe because I don't have the video upload flow enabled properly?

jesperhodge (approved):

PR #455, opened 2024-01-18
feat: insert link plugin tinymce

johnvente:
## Description 

This plugin allows you to insert a link into the text editor. Additionally, you can search for sections, subsections, and units within the course and insert the URL for any of them

### Demo

https://github.com/openedx/frontend-lib-content-components/assets/134975835/19044113-f2e3-435e-a2a9-5744716d1401

### How to test it (Devstack)

You need to have this Waffle flag activated

`new_core_editors.use_new_text_editor `

Create [waffle flag](http://localhost:18010/admin/waffle/flag/)

[Doc here](https://github.com/openedx/frontend-app-course-authoring#requirements-1)

1. Go to `~/workspace/frontend-app-course-authoring `
```
 cd ~/workspace/frontend-app-course-authoring 
```

2. Create a file called `module.config.js` with the following config

```
module.exports = {
  localModules: [
    {
      moduleName: "@edx/frontend-lib-content-components/dist",
      dir: "./frontend-lib-content-components",
      dist: "src",
    },
    {
      moduleName: "@edx/frontend-lib-content-components",
      dir: "./frontend-lib-content-components",
      dist: "src",
    },
  ],
};
```
3. Clone frontend-lib-content-components inside frontend-app-course-authoring

```
git clone -b jv/feat-tiny-mce-plugin-share-link https://github.com/eduNEXT/frontend-lib-content-components.git
cd frontend-lib-content-components && nvm use && npm install
```
If you have cloned frontend-lib-content-components already you can do this

```
cd frontend-lib-content-components && git fetch origin pull/455/head:jv/feat-tiny-mce-plugin-share-link
git checkout jv/feat-tiny-mce-plugin-share-link
nvm use && npm install
```
 
4. initialize frontend-app-course-authoring 

```
cd ~/workspace/frontend-app-course-authoring  && npm start
```


openedx-webhooks:
Thanks for the pull request, @johnvente! Please note that it may take us up to several weeks or months to complete a review and merge your PR.

Feel free to add as much of the following information to the ticket as you can:
- supporting documentation
- [Open edX discussion forum threads](https://discuss.openedx.org/)
- timeline information ("this must be merged by XX date", and why that is)
- partner information ("this is a course on edx.org")
- any other information that can help Product understand the context for the PR

All technical communication about the code itself will be done via the GitHub pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).


This is currently a draft pull request. When it is ready for our review and all tests are green, click "Ready for Review", or remove "WIP" from the title, as appropriate.

src/editors/sharedComponents/InsertLinkModal/BlocksList/index.jsx
```diff
@@ -0,0 +1,149 @@
...
+
+  const [blockState, setBlockState] = useState({
+    blockSelected: {},
+    type: blockTypes.subsection,
+    hasNavigated: false,
+    blocksNavigated: [],
+  });
+
+  const sections = getSectionsList(blocks);
+  const subSections = getChildrenFromList(
```
mariajgrimaldi, line 34:
```suggestion
  const subsections = getChildrenFromList(
```
^ Same suggestion for all the occurrences of `subSections`.

01001110J: I'm not sure about this suggestion, in Js the usual is use camelCase https://www.robinwieruch.de/javascript-naming-conventions/

I would recommend to keep the `subSections` name, or am I ignoring something?

johnvente: @01001110J You're right in JS it's better to use camelCase but "subsection" is only one world I think camelCase is usefull for two words or above, what do you think?

src/editors/sharedComponents/InsertLinkModal/BlockLink/index.test.jsx
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+import Enzyme, { shallow } from 'enzyme';
```
dcoa, line 2: Do you think we can update the test to avoid the use of Enzyme?  

According to the [OEP 0011](https://github.com/openedx/open-edx-proposals/blob/9ecc51a97d0fd6312fc47de4cd1e153f479f06a9/oeps/best-practices/oep-0011-bp-FED-technology.rst) 

>  New tests must not use Enzyme, and any repositories planning to move to React 18 or newer need to replace Enzyme.

Also, testing-library is already installed in this package.

src/editors/sharedComponents/InsertLinkModal/BlocksList/index.jsx
01001110J: As a personal recommendation I would say that it is better to use the camelCase, but the final decision is yours. :eyes:

src/editors/sharedComponents/InsertLinkModal/BlocksList/index.jsx
```diff
@@ -0,0 +1,149 @@
...
+          <Button
+            variant="tertiary"
+            className="col-1"
+            onClick={handleGoBack}
+            iconBefore={ArrowBack}
+            data-testid="block-back-navigation"
+          >
+          &nbsp;
+          </Button>
+
```
AuraAlba, line 95: Unnecessary blank line

johnvente: Removed it

mariajgrimaldi (approved):
All my comments were already addressed, thank you!

santiagosuarezedunext:
To Whom It May Concern: This functionality was approved by product in [this](https://openedx.atlassian.net/wiki/spaces/OEPM/pages/3887071233/Problem%3A+Authors+don%27t+have+enough+options+to+create+the+content+they+would+like?focusedCommentId=3932585997) post

src/editors/sharedComponents/InsertLinkModal/utils.js
```diff
@@ -0,0 +1,143 @@
...
+ * @param {string} url - The URL to be validated.
+ * @returns {boolean} - True if the URL is valid, false otherwise.
+ */
+export const isValidURL = (url) => {
+  // Regular expression for a basic URL validation
+  const urlPattern = /^(https?|ftp):\/\/[^\s/$.?#].[^\s]*$/;
+
+  // Test the provided URL against the pattern
+  return urlPattern.test(url);
+};
```
brian-smith-tcril, line 143: ### Product
Are we OK with users *needing* to add `http://` or `https://` to their links? I could imagine users being frustrated when trying to add a link using just `example.com` doesn't work. 

That being said, it seems the default link flow in TinyMCE is also somewhat odd. When testing on https://www.tiny.cloud/docs/demo/basic-example/, I found that adding `example.com` as the URL leads to the link being treated as relative. Adding `www.example.com`, however, leads to a modal

image

### Technical
I'm not sure if using regex for validation here is the right call. It should be possible to use the `URL()` constructor as a pretty much drop-in replacement https://developer.mozilla.org/en-US/docs/Web/API/URL/URL. We could also pass `type="url"` into the Paragon `Form.Control` and use `checkValidity()`. That being said, I'd be surprised if we weren't doing client-side URL validation anywhere else in the codebase. @arbrandes do you know if we have established best practices for this?

src/editors/sharedComponents/InsertLinkModal/utils.js
```diff
@@ -0,0 +1,143 @@
+import cloneDeep from 'lodash.clonedeep';
+
+export const blockTypes = {
+  section: 'chapter',
+  subsection: 'sequential',
+  unit: 'vertical',
+};
```
brian-smith-tcril, line 7: Assuming the rest of the moves I suggested in this file happen, this and `formatBlockPath` would be the only things left, at which point it might be reasonable to move each out into their own respective files (this could be in `blockTypes.js`).

That being said, we already have a file defining `blockTypes` in this repo here
https://github.com/openedx/frontend-lib-content-components/blob/f33a3b5521d709ae4b0a98a5bdd9744060177816/src/editors/data/constants/app.js#L4-L11

so it might make sense to come up with a new name for this `blockTypes`.

brian-smith-tcril (commented):
Overall this looks really great! It's super exciting to see this functionality built out.

I left a couple comments about possibly moving some things out of `utils.js`, as well as a comment with some questions about URL validation.

 Thank you so much for this PR!

src/editors/sharedComponents/TinyMceWidget/hooks.js
```diff
@@ -151,6 +154,14 @@ export const setupCustomBehavior = ({
       editor, images, setImage, openImgModal,
     }),
   });
+
+  // insert link button
+  editor.ui.registry.addButton(tinyMCE.buttons.insertLink, {
+    icon: 'link',
+    tooltip: 'Insert link',
```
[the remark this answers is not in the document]
johnvente: Yep, I agree but none of the buttons have their translations we could add the configuration like tinyMCE [docs](https://www.tiny.cloud/docs/configure/localization/):
or to use  `FormattedMessage` from i18n, what do you think?

johnvente:
> Overall this looks really great! It's super exciting to see this functionality built out.
> 
> I left a couple comments about possibly moving some things out of `utils.js`, as well as a comment with some questions about URL validation.
> 
> Thank you so much for this PR!

Hi @brian-smith-tcril! Thanks for taking a look here. I disagree with moving utils.js, principally the functions, because it's better to have functions that are reusable outside of a component. That way, you could use them in other parts of the code, and for testing purposes, it's easier to test those functions if they are outside a component. It would be tricky to test if they were inside a component, as discussed [in this article](https://medium.com/@patryk.nather/testing-local-functions-in-react-components-with-jest-55fe50a9032b). I understand that if for now they are only used inside one component, you might want to keep them there. I could move the functions to their respective component folders instead of having only one utils file, so each component will have its own utils

src/editors/sharedComponents/InsertLinkModal/utils.js
johnvente: For this case if the user is frustrated I understand that we are using this [mock up](https://www.figma.com/proto/cS42BX4OUsR8lOw6ppZHoz/Content-Libraries?type=design&node-id=1860-103852&t=6CUXwSKvDLmPMM61-0&scaling=min-zoom&page-id=1833%3A94107&starting-point-node-id=1860%3A103852&show-proto-sidebar=1) for this feature but there is not something like that, I can add it

For the URL() constructor I completely agree in use it here

src/editors/sharedComponents/TinyMceWidget/index.jsx
```diff
@@ -66,6 +69,16 @@ export const TinyMceWidget = ({
           {...imageSelection}
         />
       )}
+
+      {isInsertLinkOpen && (
+        <InsertLinkModal
+          isOpen={isInsertLinkOpen}
```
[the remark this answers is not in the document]
johnvente: it does work without `isInsertLinkOpen &&` the main problem is that the status of the modal keeps the same even if I open or close it which means for instance I selected an url that is inside the tree (an unit for example) and I open it again and it shows the link selected without that does not re-render the modal component

src/editors/sharedComponents/InsertLinkModal/utils.js
brian-smith-tcril: The linked mock specifically shows `www.example.com` (without `http://` or `https://`) as the example URL

image

In the current implementation that would fail this `isValidUrl` check.

brian-smith-tcril:
> I could move the functions to their respective component folders instead of having only one utils file, so each component will have its own utils

This sounds like a reasonable solution to me. For example, the `getChildrenFromList` method could move from `src/editors/sharedComponents/insertLinkModal/utils.js` to `src/editors/sharedComponents/insertLinkModal/BlocksList/utils/getChildrenFromList.js`, and `formatBlockPath` could move to `src/editors/sharedComponents/insertLinkModal/utils/formatBlockPath.js`

That would effectively communicate the current usage scope of each function, and it would make it simple for someone who wants to use one more generally to refactor (by simply moving the file and changing an import path)

src/editors/sharedComponents/InsertLinkModal/utils.js
johnvente: We are using a placeholder that indicates how the link should it be can I add the http:// or https:// to those urls but how could we know which one is?

johnvente: I will use http:// for those ones

brian-smith-tcril:
@johnvente it looks like the CI failures should be resolved if you rebase on latest `main` (it's `@edx` vs `@openedx` stuff that changed in https://github.com/openedx/frontend-lib-content-components/pull/457)

johnvente:
> @johnvente it looks like the CI failures should be resolved if you rebase on latest `main` (it's `@edx` vs `@openedx` stuff that changed in #457)

@brian-smith-tcril I've pulled the changes from the main branch but I'm getting this after running `npm start` 
Screen Shot 2024-02-15 at 1 27 17 PM

brian-smith-tcril:
> > @johnvente it looks like the CI failures should be resolved if you rebase on latest `main` (it's `@edx` vs `@openedx` stuff that changed in #457)
> 
> @brian-smith-tcril I've pulled the changes from the main branch but I'm getting this after running `npm start` inside course-authoring MFE Screen Shot 2024-02-15 at 1 27 17 PM

Have you tried running `npm ci` after pulling the changes?

johnvente:
> > > @johnvente it looks like the CI failures should be resolved if you rebase on latest `main` (it's `@edx` vs `@openedx` stuff that changed in #457)
> > 
> > 
> > @brian-smith-tcril I've pulled the changes from the main branch but I'm getting this after running `npm start` inside course-authoring MFE Screen Shot 2024-02-15 at 1 27 17 PM
> 
> Have you tried running `npm ci` after pulling the changes?

Yep, but same result. Even after doing it with `main` branch here I'm getting that

brian-smith-tcril:
@johnvente sorry, I missed the "inside course-authoring MFE" part. I'm working to get `frontend-app-course-authoring` updated to support `paragon` and `frontend-build` in the `@openedx` scope instead of the `@edx` scope in https://github.com/openedx/frontend-app-course-authoring/pull/818

In the meantime, you might be able to work around this error by moving `paragon` from `peerDependencies` into `dependencies`

johnvente:
@brian-smith-tcril I was able to solve it by adding "@openedx/paragon" inside the `dependencies` of `frontend-app-course-authoring` MFE. I made this PR in the meantime: https://github.com/openedx/frontend-app-course-authoring/pull/837. That works for now. Probably, you would like to remove "@edx/paragon" when the migration is finished

src/editors/sharedComponents/InsertLinkModal/utils.js
brian-smith-tcril: @felipemontoya @santiagosuarezedunext I'm hoping to get some product clarification on the "Link to URL" flow

PR comments may not be the best venue for this conversation, so I'm more than happy to find a time that works for a call if you'd rather discuss that way.

I mentioned some of it here https://github.com/openedx/frontend-lib-content-components/pull/455#discussion_r1490137186, but I was also hoping to hear some thoughts on https. 

Modern browsers use https by default to avoid making unnecessary insecure http requests. Unless a site does not support https, or a user explicitly enters `http://`, no insecure http requests are made when a user enters `example.com` into the address bar.

Many sites redirect from http to https, if you navigate to `http://github.com/`, for example, you will be redirected to `https://github.com/`. Other sites support https but do not redirect (see https://whynohttps.com/), and will simply serve the page using an insecure http connection instead (`http://example.com/` itself is an example of this).

There are also scenarios in which https is not available. If you clone this repo and serve it locally using `npm start`, it will be available at `http://localhost:8080/`, but not at `https://localhost:8080`

If we prepend `http://` to links created from urls entered without `http://` or `https://`, it is possible users will load the linked site insecurely. If we prepend `https://`, links to sites that are only served insecurely via http will be broken.

I guess I see 2 conflicting user stories here:
* As a course author (that cares about https), I want to be able to add https links without explicitly typing `https://`
* As a course author (that sometimes links to sites that don't support https), I want the links I add without explicitly typing `http://` to work.

Looking forward to hearing your thoughts!

johnvente:
Hi @brian-smith-tcril I added a translation for the tooltip, and for those URLs that are like this "[www.example.com](http://www.example.com/)," I've included a confirmation that indicates if the user wants to use "http" or "https" as a prefix. Here's an example:

https://github.com/openedx/frontend-lib-content-components/assets/134975835/27e065cb-73dd-408c-9093-bd9696dd9c49



Please let me know if the prefix is okay, or only put the translations.

I've addressed the other comments about moving utils to their respective components

santiagosuarezedunext:
[test](www.google.com)

santiagosuarezedunext:
Hi @brian-smith-tcril @johnvente I tried tools like **Notion** and **Confluence** and they prefix **http**, it is the simplest because it works in **all cases**. I think it is the responsibility of the pages to redirect to https, not our tool.
Regarding adding another message as Jhon proposes, I don't think it's a good idea. Remember that this is used by teachers who have nothing to do with technology and wouldn't know in which case to use http or htttps.
My opinion is to continue prepending http, like the rest of the tools do.

johnvente:
Hi @brian-smith-tcril I've removed the alert that I proposed [here](https://github.com/openedx/frontend-lib-content-components/pull/455#issuecomment-1949295677), but I've addressed your comments. Can you help me review it again, please?

src/editors/sharedComponents/InsertLinkModal/utils.js
```diff
@@ -0,0 +1,61 @@
...
+ */
+export const isValidURL = (url) => {
+  try {
+    // eslint-disable-next-line no-new
+    new URL(url);
+    return true;
+  } catch (error) {
+    return false;
+  }
+};
```
brian-smith-tcril, line 61: This will fail when given `www.example.com`

src/editors/sharedComponents/InsertLinkModal/index.jsx
```diff
@@ -0,0 +1,170 @@
...
+        setInvalidUrlInput(true);
+        return;
+      }
+
+      const selectedText = editor.selection.getContent({ format: 'text' });
+
+      if (selectedText.trim() !== '') {
+        const linkHtml = `<a href="${urlPath}" data-mce-href="${urlPath}" target="_blank">${selectedText}</a>`;
+        editor.selection.setContent(linkHtml);
+      }
```
brian-smith-tcril, line 70: ### Product

@felipemontoya @santiagosuarezedunext - the current implementation always shows the "Insert link" button in the toolbar

image

but this logic *requires* a user to have selected text in order to create a link. This results in a silent failure when trying to add a link without any selected text.

### Techinical

@johnvente I also found what feels like a bug here. If you create a new at first everything seems fine

image

image

but then, if you select the link text and click "Insert link" again you're shown a new modal with no URL. Entering a URL and clicking save results in nested links

image

brian-smith-tcril (commented):
I left a couple more comments. The biggest ones are in regard to this plugin only supporting adding links to text that has been selected, and behaving unexpectedly if the user tries to add a link without selecting text or edit a link.

src/editors/sharedComponents/InsertLinkModal/utils.js
johnvente: That's right because doesn't have the protocols, the user should put the full url

brian-smith-tcril: @santiagosuarezedunext please correct me if I'm misunderstanding, but when you recommended we prepend `http://`you were saying "if the user enters `www.example.com` the link should become `http://www.example.com`," right?

johnvente:
Hi @brian-smith-tcril thanks for taking a look over here again! I will be checking the bugs found

src/editors/sharedComponents/TinyMceWidget/pluginConfig.js
```diff
@@ -53,9 +53,10 @@ const pluginConfig = ({ isLibrary, placeholder, editorType }) => {
           buttons.outdent,
           buttons.indent,
         ],
-        [imageUploadButton, buttons.link, buttons.unlink, buttons.blockQuote, buttons.codeBlock],
```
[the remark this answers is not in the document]
johnvente: @santiagosuarezedunext Just to clarify we should leave the link plugin that is currently and limit this one only to add new links? becase the scope change a lot if not

johnvente:
Hi @brian-smith-tcril  thanks for reviewing this. Now we've limited the "jum to" links I have left the plugin used default  by tinyMCE to insert links and this one now will be added only for add links of the sections of the course I updated the demo video of the PR to see how looks. Now you can do these following things:

1. Add a link of section, subsection and so on.
2. Edit links added as well
3. I'v checked the bug you mentioned [here](https://github.com/openedx/frontend-lib-content-components/pull/455#discussion_r1496227991)


Thanks again for the review, could you review agin this please?

brian-smith-tcril:
@santiagosuarezedunext I have a few more product questions here:

* How should we handle hover text> It is currently very similar for both link widgets - I assume we want to differentiate these.
image
image

* How should we handle adding links without selected text?
The built-in link editor autopopulates the "text to display" field with the selected text.
image
The section/subsection/etc. link modal shows a "The text for the URL was not selected" error
image
And after selecting a section/subsection etc., the error is not shown but the "Save" button is disabled
image

santiagosuarezedunext:
@brian-smith-tcril That's right, the hover text of the new button should say "Jump to course section".
Regarding the message, @johnvente let it remain visible in the last section of the modal. 😊👍

johnvente:
Hi @brian-smith-tcril @santiagosuarezedunext I've changed the title of the hover to 

"Jump to course section"
Screen Shot 2024-02-29 at 12 36 01 PM


For handling the text not selected, I've disabled the buttons for course sections. That way, the user won't be able to select any of them, and the error is shown

Screen Shot 2024-02-29 at 12 36 12 PM

johnvente:
Hi there @brian-smith-tcril! just a quick reminder over here could check this again please?

brian-smith-tcril:
Found another edge case:
* Create a new external link
image
* Select the link and open "jump to course section," save an internal link
image
* Open the external link editor, see the new URL
image
* Change the url back to an external link
image
* Select the text and open "jump to course section" again - the link will show it's going to an internal page but the url is an external link
image
* Click the "remove link" button and hit save
image
* See the external link still exists
image

santiagosuarezedunext:
I reference the discussion we had in Slack about changes to this functionality: https://openedx.slack.com/archives/C05714Z26N9/p1709832678914629

@johnvente, please help me by disabling the icon if no text has been selected

Thank you all very much for your patience and your help in carrying out this functionality, it will help the teachers a lot.
CC: @brian-smith-tcril

src/editors/sharedComponents/TinyMceWidget/hooks.js
```diff
@@ -151,6 +156,14 @@ export const setupCustomBehavior = ({
       editor, images, setImage, openImgModal,
     }),
   });
+
+  // insert link button
+  editor.ui.registry.addButton(tinyMCE.buttons.insertLink, {
+    icon: 'new-tab',
+    tooltip: translations?.insertLinkTooltipTitle ?? '',
+    onAction: openInsertLinkModal,
+  });
```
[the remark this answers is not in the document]
johnvente: Thanks! I added it and works perfectly

johnvente:
Hi @brian-smith-tcril sorry for the delay here I resolved the edge case that you found [here](https://github.com/openedx/frontend-lib-content-components/pull/455#issuecomment-1981879544) and now the tooltip for the 'jum to' is disabled when there is not a selected text, could check again this please? Thanks a lot

johnvente:
> Hi @brian-smith-tcril sorry for the delay here I resolved the edge case that you found [here](https://github.com/openedx/frontend-lib-content-components/pull/455#issuecomment-1981879544) and now the tooltip for the 'jum to' is disabled when there is not a selected text, could check again this please? Thanks a lot

Hi @brian-smith-tcril a reminder over here 👀

brian-smith-tcril:
@johnvente I'm planning on testing/reviewing this again either later this afternoon or tomorrow.

brian-smith-tcril:
I would like to verify a bit of intended behavior here:

### Scenario 1
* Create a "jump to page link"
* Place cursor/insertion bar inside link text
image
* See the built-in link button in "edit" mode, but the "jump to" link button greyed out
image
  * clicking on it will show  
image

This makes sense from a code perspective considering the logic for enabling the button is simply checking to see if we have a selection, but it does feel a bit odd.

### Scenario 2
* Create a "jump to" link
* Select the entire link
image
* Click the "jump to" link button
* See the top of the navigation tree (no indication that something is already linked)
image

## Overall thoughts

The root of these UX oddities is the fact that the built-in link button is both an "add link" button and a "edit link" button, while the "jump to" link button is only an "add link" button (although it will properly overwrite existing links if the entire link text is selected before opening the dialog)

@jmakowski1123 are we OK with shipping this without any *edit* "jump to" link functionality?

jmakowski1123:
> I would like to verify a bit of intended behavior here:
> 
> ### Scenario 1
> * Create a "jump to page link"
> * Place cursor/insertion bar inside link text
>   image
> * See the built-in link button in "edit" mode, but the "jump to" link button greyed out
>   image
>   
>   * clicking on it will show
>     image
> 
> This makes sense from a code perspective considering the logic for enabling the button is simply checking to see if we have a selection, but it does feel a bit odd.
> 
> ### Scenario 2
> * Create a "jump to" link
> * Select the entire link
>   image
> * Click the "jump to" link button
> * See the top of the navigation tree (no indication that something is already linked)
>   image
> 
> ## Overall thoughts
> The root of these UX oddities is the fact that the built-in link button is both an "add link" button and a "edit link" button, while the "jump to" link button is only an "add link" button (although it will properly overwrite existing links if the entire link text is selected before opening the dialog)
> 
> @jmakowski1123 are we OK with shipping this without any _edit_ "jump to" link functionality?

It feels odd to me that we wouldn't mirror the range of actions possible with both the native link workflow and the jump-to link workflow. Is there significant additional lift to add editing functionality to the jump-to link?

src/editors/sharedComponents/TinyMceWidget/hooks.js
```diff
@@ -151,6 +156,19 @@ export const setupCustomBehavior = ({
...
   });
+
+  // insert link button
+  editor.ui.registry.addButton(tinyMCE.buttons.insertLink, {
+    icon: 'new-tab',
+    tooltip: translations?.insertLinkTooltipTitle ?? '',
+    onAction: openInsertLinkModal,
+    onSetup(api) {
+      editor.on('SelectionChange', () => {
+        api.setDisabled(editor.selection.getContent().length === 0);
```
brian-smith-tcril, line 167:
```suggestion
        api.setDisabled(
          editor.selection.getContent().length === 0 &&
          editor.dom.getParent(editor.selection.getNode(), 'a[href]') === null
        );
```

This suggestion does not include logic to check if the link is an internal "jump to" link or an external link. I think it should be possible to create a helper function with that logic and utilize it both here and when getting the block id the `useEffect()` in `InsertLinkModal/index.jsx`

src/editors/sharedComponents/InsertLinkModal/index.jsx
```diff
@@ -0,0 +1,220 @@
...
+      const regexDataBlockId = /data-block-id="([^"]+)"/;
+      const regexHref = /href="([^"]+)"/;
+      const matchDataBlockId = selectedHTML.match(regexDataBlockId);
+      const matchHreUrl = selectedHTML.match(regexHref);
+
+      // Extracting the value from the match
+      const dataBlockId = matchDataBlockId ? matchDataBlockId[1] : null;
+      const hrefUrl = matchHreUrl ? matchHreUrl[1] : null;
+      const blockSelectedUrl = selectedBlocks?.[dataBlockId]?.lmsWebUrl;
+      const hasExternalUrl = hrefUrl !== blockSelectedUrl;
```
brian-smith-tcril, line 122: If I'm understanding this correctly it seems this is trying to find the block associated with the selected link. This was not working properly for me when selecting the entirety of a link's text, or when clicking inside a link after making the changes I suggested to `hooks.js`

brian-smith-tcril (commented):
@jmakowski1123 re:
> Is there significant additional lift to add editing functionality to the jump-to link?

It seems there is logic in place that is attempting to provide "edit link" functionality that:
* Isn't currently functioning as intended (I didn't work at all in my testing)
* Isn't accessible with the current "disable button" logic

I left a few comments here, one pointing out the section that seems to not be properly getting the linked section, and one with a suggestion that should move the disable button logic in the right direction (but not a complete solution).

I wouldn't expect this to be a huge lift, it's more bug fix than new feature.

johnvente:
I won't be able to review this in a week, so I will check this on April 1st

jmakowski1123:
If it's a bug fix, then we definitely should not release this without it, or should have a plan in place to fast-follow on the fix.

santiagosuarezedunext:
@johnvente I agree with @brian-smith-tcril and @jmakowski1123, it is necessary that when you click on a text that already has a "jump to link" assigned, it should be shown when you go to edit or overwrite it.

src/editors/sharedComponents/TinyMceWidget/hooks.js
johnvente: This line is not working for me, but I found the part to check if a link has already added
`editor.dom.getParent(editor.selection.getNode(), 'a[href]') === null`

src/editors/sharedComponents/InsertLinkModal/index.jsx
johnvente: Yes, this is trying to validate if there is a external link or has jump to link selected, but in a logic below it checks if has a parent node and I was not having account that so that was the problem

johnvente:
Hi there @brian-smith-tcril I added the fix for editing a text that has already a "jump to link" please let me know if it works as expected I will be checking it

src/editors/sharedComponents/InsertLinkModal/api.js
```diff
@@ -0,0 +1,29 @@
...
+    );
+
+    const { blocks } = data;
+    const blocksFormat = Object.keys(blocks).reduce(
+      (prevBlocks, key) => ({
+        ...prevBlocks,
+        [key]: camelCaseObject(blocks[key]),
+      }),
+      {},
+    );
```
johnvente, line 21: All object in blocks response come in `snake_case` this converts all of them to `camelCase` for avoiding problems in when we need this data

brian-smith-tcril:
> Hi @brian-smith-tcril I added the fix for editing a text that has already a "jump to link" please let me know if it works as expected I will be checking it

It works when selecting the link text but not when the insertion bar is in the link text

image

image

image

you can see the default add/edit link button is highlighted when the insertion bar is in the link text, and the remove link button is enabled as well, but the jump to button is still disabled

johnvente:
> > Hi @brian-smith-tcril I added the fix for editing a text that has already a "jump to link" please let me know if it works as expected I will be checking it
> 
> It works when selecting the link text but not when the insertion bar is in the link text
> 
> image
> 
> image
> 
> image
> 
> you can see the default add/edit link button is highlighted when the insertion bar is in the link text, and the remove link button is enabled as well, but the jump to button is still disabled

Yep, but why to check the insertion bar? as far I knew we wanted to have it only when the text is selected, we have short time for this I'm not sure how long take to check the insertion bar

brian-smith-tcril:
> Yep, but why to check the insertion bar?

Because that's how the native link button behaves. It doesn't make sense to have different UX for the internal link button and the native link button.

johnvente:
> > Yep, but why to check the insertion bar?
> 
> Because that's how the native link button behaves. It doesn't make sense to have different UX for the internal link button and the native link button.

Okay, I will check that would it be enough with that?

johnvente:
> I found another issue.

@brian-smith-tcril I added a fix for that use case but for the insertion bar not yet since the native plugin is active even when there is not text selected or when the editor is empty completely so is active all the time we should limit our plugin only for text selected instead of trying to have the behavior of the native plugin I'm not sure how long can take and I won't be able to check it after this friday please let me know what do you recommend here

brian-smith-tcril:
> since the native plugin is active even when there is not text selected

The grey indicates it will edit the link instead of create a new one

image

image

vs 

image

image

>  I won't be able to check it after this friday 

I'm more than happy to take this over and get it across the finish line.

johnvente:
> I'm more than happy to take this over and get it across the finish line.

That is more like new feature that a bug fixing since our code was made for text selected not when the user has insertion bar to edit a link. I will try as much as I can to check it but the time is very short

johnvente:
Hi @brian-smith-tcril I added the support to edit a new link as you [mentioned here](https://github.com/openedx/frontend-lib-content-components/pull/455#issuecomment-2033367689) could you take a look please? I will be checking  it

brian-smith-tcril:
Found a couple more issues

* Use "jump to" button to create a link
* Insertion bar and selection is now completely gone from the editor
* Button state is now both disabled and has the grey box
image

compared to

* Use native link button to create link
* Selection stays, insertion bar stays, native link button state updated correctly

A similar thing with the button state happens when deleting a link using the jump to button

Also

* Create link using native link editor (to an external site)
* Select the link (or place insertion bar in it)
* See "jump to" button in edit state
image
* Click on "jump to" button
* No page is navigated to (since it's an external link)

And another one that's a bit more complicated

* Create a link using the native link editor (to page url copied by opening a "jump to" link in the native editor)
* Select the link (or place insertion bar in it)
* Clock on the "jump to" button
* See no page selected in the jump to menu

I realize this last one is because the url isn't being parsed to determine if the link is internal or not, but instead an id is being associated with it - but it leads to confusing UX.

johnvente:
Sorry for the delay here, I won't be able to make changes here but any request please let know to @felipemontoya @mariajgrimaldi 

Good luck

brian-smith-tcril:
@johnvente thank you for all of your work on this! I plan on taking this over at the beginning of next week. I know it's turned out to be more complicated than expected, and you've done an amazing job getting it to this point. I'm really happy to be able to use this as a base to get it across the finish line.

Thanks again and all the best!

src/editors/sharedComponents/TinyMceWidget/hooks.js
```diff
@@ -151,6 +156,28 @@ export const setupCustomBehavior = ({
...
+    tooltip: translations?.insertLinkTooltipTitle ?? '',
+    onAction: openInsertLinkModal,
+    onSetup(api) {
+      editor.on('SelectionChange', () => {
+        const node = editor.selection.getNode();
+        const isLink = node.nodeName === 'A';
+        const hasTextSelected = editor.selection.getContent().length > 0;
+        api.setActive(isLink);
+        api.setDisabled(!isLink && !hasTextSelected);
+      });
```
brian-smith-tcril, line 177:
```suggestion
      editor.on('SelectionChange', () => {
        const node = editor.selection.getNode();
        
        const isLink = node.nodeName === 'A';
        let isInternalLink = false;
        if (isLink) {
          const selectedHTML = editor.selection.getContent({ format: 'html' }) || node.outerHTML;
          const regexDataBlockId = new RegExp(/data-block-id="([^"]+)"/);
          isInternalLink = regexDataBlockId.test(selectedHTML);
        }

        const hasTextSelected = editor.selection.getContent().length > 0;
        api.setActive(isInternalLink);
        api.setDisabled(!isInternalLink && !hasTextSelected);
      });
```

By doing this I was able to get it so the "jump to" button is only enabled when trying to edit internal links, but I was able to break it decently quickly (and uncovered another bug in the process).

The native link editor doesn't remove `data-block-id`, so by
* Creating an internal link using the "jump to" dialog
* Editing that internal link and changing it to be an external link using the native link dialog

We end up with a link on the page with an associated block id even though the url doesn't match anymore.

This gets cleaned up when opening the "jump to" menu during the block id/url validation, but I'd like to clean it up before getting in there.

I'm looking into ways to do that without duplicating too much code.

santiagosuarezedunext:
Hi @brian-smith-tcril, I hope you're doing well. I wanted to ask how the development of this feature is going. What is still needed to finish closing it? Do you need help with anything?

Best regards

brian-smith-tcril:
@santiagosuarezedunext I have been focused on the edx-platform node 18 upgrade. I'll get back to this once I've worked through the issues with that.

santiagosuarezedunext:
Hello @brian-smith-tcril, I hope you are well.
I wrote to you to follow up on the functionality. Is there any update on this? What is needed to finish it? Please tell me if anything is needed from our side.

brian-smith-tcril:
@santiagosuarezedunext I have been busy with Redwood and haven't had time to dive into this, but I have deployed the current code from this PR on a sandbox here: https://studio.pr-1070-e8373e.sandboxes.opencraft.hosting/

The current implementation differs significantly from what was originally proposed and approved here: https://openedx.atlassian.net/wiki/spaces/OEPM/pages/3887071233/Problem+Authors+don+t+have+enough+options+to+create+the+content+they+would+like?focusedCommentId=3932585997

To move forward with this, the implementation deployed to the sandbox should go through Product and UX/UI testing. That testing should result in updated requirements for getting this PR into a mergeable state.

If I were to perform the Product/UX/UI testing, I would not pass the current implementation because:
* There should not be 2 "add link" buttons, adding external and internal "jump to" links should be possible using the same button

If, however, someone decides that having 2 link buttons is acceptable, I would still not pass the current implementation because:
* There should always be a way to add a link without selecting text
  * This is currently only possible with external links
* The internal "jump to" link button should not allow opening the modal to try to edit external links
  * The internal "jump to" link button currently opens to "search course pages" when an external link is opened with it. 
* Adding a link to an internal page by pasting a URL into the external link editor, then opening that link in the internal "jump to" editor, should result in the correct internal page being selected
  * The internal "jump to" link button currently opens to "search course pages" when an internal link added via URL using the external link editor is opened with it.
* The internal "jump to" link button is currently using an icon associated with "open in new tab"

@ali-hugo tagging you for visibility. I'm going to put this PR in draft mode until Product and UI has determined the best path forward.

ali-hugo:
@brian-smith-tcril Thanks for putting this on my radar. 

I wanted to check out the sandbox so created an account, but I don't seem to be receiving a verification mail. Is it possible for you to verify my email address on your side?

santiagosuarezedunext:
@brian-smith-tcril i have the same error that @ali-hugo have.

brian-smith-tcril:
@santiagosuarezedunext I have marked your user as "Active" and "Staff." Could you try again and let me know if you're still having problems logging in?

Thank you for your patience!

santiagosuarezedunext:
It already works for me, thanks @brian-smith-tcril

KristinAoki:
Hi! Thank you for your contribution to this repo. This repo is being deprecated and all the code is being moved to `frontend-app-course-authoring`. If you would like your work to be included in the transition please resolve outstanding requests for this PR. This PR will automatically close once `frontend-app-course-authoring` PR [#1208](https://github.com/openedx/frontend-app-course-authoring/pull/1208) is merged.

bradenmacdonald:
Closing as we've now moved this code, per the notice above. Please re-open this PR against https://github.com/openedx/frontend-app-course-authoring/ if you'd like to continue it. Feel free to reach out to me if you aren't sure how to do that using git and I can help (basically: add your local `content-components` repo as a remote in `course-authoring`, then pull the branch and rebase it).

openedx-webhooks:
@johnvente Even though your pull request wasn’t merged, please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/frontend-lib-content-components&entry.867055334=https://github.com/openedx/frontend-lib-content-components/pull/455&entry.1484655318=https://github.com/johnvente&entry.752974735=2024-01-18+02:37&entry.1917517419=2024-08-29+19:07&entry.2133058324=No) so we can improve your experience in the future.

PR #456, opened 2024-01-26
feat: bumped frontend-platform to v6

BilalQamar95:
**Description:**
- Major version upgrade for `frontend-platform`
- Updated snapshots & fixed failing tests

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -46,8 +46,8 @@
       },
       "devDependencies": {
         "@edx/browserslist-config": "^1.1.1",
-        "@edx/frontend-build": "13.0.1",
-        "@edx/frontend-platform": "5.5.4",
+        "@edx/frontend-build": "13.0.14",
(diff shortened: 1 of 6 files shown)


brian-smith-tcril (commented):
This is only updating `frontend-platform` in `devDependencies`, it will need to be updated in `peerDependencies` too.

https://github.com/openedx/frontend-lib-content-components/blob/fcc9a054ce5dd39f3bb12555066388194bd468b6/package.json#L100

We also don't have anything in place to test that this will still work with `frontend-platform` versions < 6, so it may need to be done as a breaking change.

If that's the case, it might make sense to wait for https://github.com/openedx/frontend-lib-content-components/pull/457 to make the breaking change commit so we don't end up with 2 major version bumps back to back.

BilalQamar95:
Closing since package upgrades of this PR were already taken care of as part of [#457](https://github.com/openedx/frontend-lib-content-components/pull/457)

PR #459, opened 2024-03-07
fix: problem type not selectable

jesperhodge:
JIRA: https://2u-internal.atlassian.net/browse/TNL-11465

This is a problem we're experiencing on stage due to a bug in paragon. This is intended as a temporary fix until we can dig into why this is happening in paragon. But basically the Context for the `SelectableBox` is missing a provider in paragon. The short-term fix is to copy over paragon's selectablebox component and fix it. This is done so that our quick fix doesn't break anything else in paragon for now or cause any unexpected issues.

(locally the problem didn't pop up until we copied over the paragon component. Theory is that locally a different paragon version was installed for some reason.)


rayzhou-bit (approved):
nit: Could you add a comment to `ProblemTypeSelect.jsx` on how this is a 'fixed' version of the paragon component?
Otherwise, I tested this change and it looks good to me!

PR #465, opened 2024-03-15
fix: remove useEffect that fires before search params

KristinAoki:
JIRA Ticket: [TNL-11505](https://2u-internal.atlassian.net/browse/TNL-11505)

> It was reported an issue regarding videos not populating the Video ID field when uploading a video via the video component. This issue results in the display of an error message stating, “No playable video source found,” due to the absence of the video ID in the designated field. A workaround for this video issue is to manually copy the video ID from the video block and paste it into the Video ID field.

Testing
1. Open a course.
2. Edit to the video component.
3. Click "Replace video"
4. Select the desired video for the unit.
5. The video will upload with the Video ID

### DIFF
--- a/src/editors/containers/VideoEditor/components/VideoEditorModal.jsx
+++ b/src/editors/containers/VideoEditor/components/VideoEditorModal.jsx
@@ -5,7 +5,6 @@ import * as appHooks from '../../../hooks';
 import { thunkActions, selectors } from '../../../data/redux';
 import VideoSettingsModal from './VideoSettingsModal';
 // import SelectVideoModal from './SelectVideoModal';
-import * as module from './VideoEditorModal';
 
 export const {
   navigateTo,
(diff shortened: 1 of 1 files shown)


src/editors/containers/VideoEditor/components/VideoEditorModal.jsx
```diff
@@ -14,9 +13,7 @@ export const {
 export const hooks = {
   initialize: (dispatch, selectedVideoId, selectedVideoUrl) => {
     // eslint-disable-next-line react-hooks/rules-of-hooks
-    React.useEffect(() => {
-      dispatch(thunkActions.video.loadVideoData(selectedVideoId, selectedVideoUrl));
-    }, []);
+    dispatch(thunkActions.video.loadVideoData(selectedVideoId, selectedVideoUrl));
```
rayzhou-bit, line 16: Nit:
The `rules-of-hooks` disable in L15 can be removed.
If returnToGallery is renamed to useReturnToGallery, the other `rules-of-hooks` disables can be removed as well.

KristinAoki:
> Was this error a regression in functionality? if so, do we want to add a test for it?

This was not an regression in functionality. There bug has been around for a while and was not consistently reproducible. I do not think any additional tests are needed.

connorhaugh (approved):

PR #477, opened 2024-05-20
fix: video selection gallery scroll

KristinAoki:
JIRA Ticket: [TNL-11615](https://2u-internal.atlassian.net/browse/TNL-11615)
> Video selection gallery in video editor is not scrollable

The default scrolling for the full screen modal was disabled and `<Scrollable/>` component was added inside the modal. however, the modal content scrollbar was still disabled. This PR moves the `<Scrollable/>` component as a wrapper of the modal body. This change impacts the video selection gallery, image selection and edit modal, and source code editor modal.

Testing
1. Open a html block
2. Click on the Image button in the toolbar
3. Check that you can scroll through the loaded images
4. Click on an image to add
5. Check that everything is function in the image edit screen.
6. Click "Cancel"
7. Click on the HTML button in the toolbar
8. Add a bunch of text
9. Confirm that the modal expands to its max height then adds scroll bar if it exceed screen height
10. Exit out of html editor
11. Open a video block
12. Click "Replace video"
13. Check that you can scroll through the videos


src/editors/sharedComponents/SelectionModal/index.jsx
```diff
@@ -81,7 +81,7 @@ export const SelectionModal = ({
         </Button>
       )}
       title={intl.formatMessage(titleMsg)}
-      bodyStyle={{ background, padding: '3px 24px' }}
+      bodyStyle={{ background }}
```
rayzhou-bit, line 84: nit: I'm not seeing padding between the bottom selection and the bottom after scrolling down. This isn't blocking any information or selection.

rayzhou-bit (approved):
I tested this and it looks good! Got one nit that doesn't need to be addressed.

PR #479, opened 2024-05-29
fix: removal of content after problem type tags

KristinAoki:
JIRA Ticket: [TNL-10674](https://2u-internal.atlassian.net/browse/TNL-10674)
> Reporting that a pre-existing problem containing a linked image is missing when they go to edit the problem in the visual editor, and also missing if they switch to the advanced editor.

This PR fixes the intermittent problem where there is OLX after the problem type tags, (`multiplechoiceresponse`, `numericalresponse`, etc.) for simple problems. Currently the parser will some times put the OLX inside the question section; other times the parser will remove the OLX. When the OLX is removed, the user switches to the advanced editor expecting to see it, but it is not present. This PR updates the parser to check if there is OLX content after the problem type tags, and if found, redirect to the advanced editor. This change removes the possibility that previous OLX will be blindly removed and that content will be rearranged blindly.

Testing

1. Create a new  problem block
2. Choose one of the simple problem types
3. Add content to the problem
4. Click "Show advanced settings"
5. Click "Switch to advanced editor"
6. Before the closing problem tag (`</problem>`) add `<p>content after problem type tags</p>`
7. Click "Save"
8. Click "Edit" for the problem that you just created
9. Should automatically open the problem in the advanced editor
10. Remove `<p>content after problem type tags</p>`
11. Click "Save"
12. Click "Edit" for the problem that you just created
13. Should automatically open the problem in the visual editor

### DIFF
--- a/src/editors/containers/ProblemEditor/data/OLXParser.js
+++ b/src/editors/containers/ProblemEditor/data/OLXParser.js
@@ -491,6 +491,42 @@ export class OLXParser {
     return res;
   }
 
+  /** checkTextAfterProblemTypeTag(problemType)
+   * checkTextAfterProblemTypeTag takes a problemType. The problem type is used to determine
+   * if there is olx after the answer choices the problem. Simple problems are not expected
+   * to have olx after the answer choices and returns null. In the event that a problem has
+   * olx after the answer choices and error is raised resulting in a redirection to the
+   * advanced editor.
+   * @param {string} problemType - string of the olx problem type
+   * @return {null}
+   */
+  checkTextAfterProblemTypeTag(problemType) {
+    let problemTagIndex = this.richTextProblem.length - 1;
+    Object.entries(this.richTextProblem).forEach(([i, value]) => {
+      if (Object.keys(value).includes(problemType)) {
+        problemTagIndex = i;
+      }
+    });
+
+    // The richTextProblem array might have a new line character as the final element and the problem tag
+    // is the second to last element. The new line character should not impact the need for the advanced
+    // editor open so the index is checked against the length of the richTextProblem array minus 2.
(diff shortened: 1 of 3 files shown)


src/editors/containers/ProblemEditor/data/OLXParser.js
```diff
@@ -634,14 +670,16 @@ export class OLXParser {
       throw new Error('Misc Attributes asscoiated with problem, opening in advanced editor');
     }
 
+    const problemType = this.getProblemType();
+
+    this.checkTextAfterProblemTypeTag(problemType);
```
jesperhodge, line 675: When reading the code, it's not immediately obvious to me that this function can throw an error that errors out of the running code. What do you think about making it explicit, like in the line above with `throw new Error('Misc Attributes asscoiated with problem, opening in advanced editor');`?

For example, since the error message here is the same whether you have invalid keys or invalid tags, the `checkTextAfterProblemTypeTag` can just be something like `hasXMLAfterProblemTypeTag` and then here you could do
```
if (this.hasXMLAfterProblemTypeTag(problemType) {
  throw new Error(`OLX found after the ${problemType} tags, opening in advanced editor`);
}
```
. The benefit of doing it in a similar way to what I suggested is that any reader of the code can immediately see that an error is thrown, which makes the code clearer and the logic less confusing in my opinion.

jesperhodge (changes requested):
I love it!

Great fix.
I just added a little stylistic comment and after we resolve that discussion I'll approve.

KristinAoki:
@jesperhodge I think your proposed solution makes sense. I think that is what we were trying to do initially, but as new visual editor exceptions appears, we re-oriented the code to look for the exceptions.

jesperhodge (approved):

Issue #480, opened 2024-05-31
Unlimited attempts cannot be updated in problem editor


mariajgrimaldi:
### Expected behavior

As a course author, I want to set up a problem with unlimited attempts, so that students can submit their answers as many times as they want.

### Actual behavior

As a course author, I cannot assign a problem with unlimited attempts. It happens when you already have a set number of attempts and want to change to unlimited attempts.

### Steps to reproduce

1. From **Studio** create a problem.
2. Configure the problem with a number of attempts, e.g.: `5`, and save.
3. Edit the problem try to assign unlimited attempts, and save
4. The problem continues to have the same number of attempts assigned previously

https://github.com/openedx/wg-build-test-release/assets/64033729/f722b3a2-ed43-4081-ac13-dd84c7008291

### Additional information

It happens in all problems, but it does not happen with the legacy editor.

Original issue: https://github.com/openedx/wg-build-test-release/issues/371

mariajgrimaldi:
Here's what I've found:
- This happens even without publishing the unit
- When selecting unlimited attempts after setting a limited attempts to any number, the request doesn't include the unlimited attempt change:
![image](https://github.com/openedx/wg-build-test-release/assets/64440265/860b6da1-f3c0-4545-b06b-dcb6572d5292)
- When changing the number of attempts, it sends the new change accordingly:
![image](https://github.com/openedx/wg-build-test-release/assets/64440265/ddfb8577-ed47-47ec-8b8a-a663ec96b75a)

I'll be reporting this in the frontend-lib-content-components repo.

mariajgrimaldi:
FYI @arbrandes @brian-smith-tcril @openedx/2u-tnl, please tag anyone who can help.

brian-smith-tcril:
in some quick testing on the sandbox using the network tab in the browser inspector (filtering to `method:POST`) I was able to find that checking "unlimited attempts" sends nothing for `max_attempts` in the `metadata` part of the payload. By manually entering `0` for number of attempts I was able to get the "You have used 0 of x attempts" message to go away.

mariajgrimaldi:
@brian-smith-tcril: Exactly, thanks for testing it out! I mentioned something similar in the comment I left [here](https://github.com/openedx/frontend-lib-content-components/issues/480#issuecomment-2141201587). However, I'm not sure why setting unlimited attempts first works sending the correct metadata but not changing it to unlimited.

brian-smith-tcril:
[...]

PR #483, opened 2024-06-04
fix: parser not saving unlimited attempts

KristinAoki:
## Description
This PR fixes problems where the maximum attempts is null by default. Previously, when a course had unlimited attempts by default and a user created a new problem, the unlimited option was not checked so the scoring card said, "1 point • 0 attempts". Now for new problems the unlimited option is checked and the card reads, "1 point • Unlimited attempts". This PR allows resolved the issue when changing a problems number of attempts from unlimited to a number and back to unlimited, the revert back to unlimited was not saved. This is because the default value of settings is not saved to problems so that it is able to automatically update if the default settings are changed. However, not saving the unlimited value did not revert the problem back to use the default maximum attempts. Now the unlimited value is saved even when it is the default value to reset the problem. The backend value is null so it does not inhibit the updating of the problem dynamically to match the advanced setting changes.

## Supporting information

Github Issue #480 

## Testing

1. Open a course
2. Navigate to the Advanced settings page and confirm that "Maximum attempts" is set to `null`
3. Navigate to a unit page
4. Add a new simple problem
5. Check that the scoring card reads, "1 point • Unlimited attempts"
6. Click on the scoring card
7. Confirm that the unlimited checkbox is checked
8. Uncheck the box and add a new number of attempts
9. Click "Save"
10. Click "Edit" for the problem you just saved
11. Click on the scoring card
12. Check the unlimited checkbox
13. Click "Save"
14. Should not mention the number of attempts in the problem preview
15. Navigate to the Advanced settings page
16. Set "Maximum attempts" is set to 5
17. Navigate back to unit page
18. Confirm that problem preview shows the text "You have used 0 of 5 attempts"

### DIFF
--- a/src/editors/containers/ProblemEditor/components/EditProblemView/hooks.test.js
+++ b/src/editors/containers/ProblemEditor/components/EditProblemView/hooks.test.js
@@ -305,7 +305,7 @@ describe('EditProblemView hooks parseState', () => {
       });
     });
 
-    it('returned parseState content.settings should not include default values', () => {
+    it('returned parseState content.settings should not include default values (not including maxAttempts)', () => {
       const problem = {
         ...problemState,
         problemType: ProblemTypeKeys.NUMERIC,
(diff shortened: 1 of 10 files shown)


jesperhodge (approved):
Awesome!

PR #489, opened 2024-06-26
fix: simple editor without solution not loading

KristinAoki:
This PR resolves the problem editor not load new simple problems and reimplements the changes in #484 

Testing

1. Create a new problem
2. Select a problem type
3. Confirm that the editor loads
4. Refresh the page
5. Repeat 2 through 4 until all simple problems have been tested


jesperhodge (approved):
Very excited about the performance improvement! That's just a great feature

PR #492, opened 2024-07-09
fix: text editor opening blank with no images

KristinAoki:
## Description

This PR fixes issues the following issues introduced by #490:
 - When no images or no static images need to be changed, the text and problem editors did not load previously saved content
 - When saving images with the raw text editor and advanced problem editor, the images when not load and show the broken  src icon. After re-opening the editor, the src would be incorrectly formatted and not show the original static url

### Testing raw editor
1. Open an existing raw html block
2. Confirm that the html loads 
3. Add an image using the "Studio URL"
4. Add an image using the "Web URL"
5. Click "Save"
6. Confirm that the images are viewable on the unit page
7. Re-open the block
8. Confirm the image srcs are the same as originally saved
9. Click "Save"
10. Confirm that the images are still viewable
11. Add a new raw html block
12. Confirm that the default content loads

### Testing text editor
1. Open an existing html block with no images
2. Confirm that text loads
3. Click "Cancel"
4. Open an existing html block with images
5. Confirm that the text and images load
6. Click "Save"
7. Confirm that the images load in block preview on the unit page
8. Add a new text block
9. Confirm that the editor loads

### Testing problem editor
1. Open an existing problem block with no images
2. Confirm that problem loads
3. Click "Cancel"
4. Open an problem html block with images
5. Confirm that the problem and images load
6. Switch to the advanced editor
7. Confirm that all relative paths are use `/static/`
8. Click "Save"
9. Confirm that the images load in block preview on the unit page
10. Add a new problem block
11. Select any of the basic problems
12. Confirm that the editor loads


rayzhou-bit (approved):
lgtm! I tested with inserting tables as well which was also an issue before.

PR #500, opened 2024-08-02
fix: remove imports breaking build

KristinAoki:
This PR removes paragon scss variable imports from `.scss` files. These imports break the build of the course authoring MFE because of a `ModuleErrorBuild`. The only variable used is `$gray-500` and the hex code for the variable is the same in the openedX and edX themes, `#707070`, so replacement will have no affect on the UI


rayzhou-bit (approved):

bradenmacdonald:
If you prefer to keep the variables, I think https://github.com/open-craft/frontend-lib-content-components/commit/5ff22c1ee6cdbbb98d8578a1313ddb08cbd4b873 would fix the error; assuming it's the same as a similar error I recently saw. This fix seems fine too though.

PR #497, opened 2024-07-31
fix: setAssetToStaticUrl regex matcher

KristinAoki:
## Description

This PR fixes the error of relative asset urls incorrectly rewriting the conversion to the static url.  The regex matcher was catching all relative asset url formats:

- `/assets/courseware/v1/hash/asset-v1:org+num+run+type@asset+block/image.jpeg`
- `/asset-v1:org+num+run+type@asset+block@image.jpeg`
- `/asset-v1:org+num+run+type@asset+block/image.jpeg`

This resulted in the first two version rewriting correctly as `/static/image.jpeg` and the third rewriting incorrectly as `/static/assetblockimage.jpeg`.

## Testing

1. Open a raw editor
2. Add an image with a static link
3. Save the block
4. Using browser dev tools, copy the link for the image
5. Reopen the block
6. Confirm that the static is unaltered
7. Paste the previously copied image
8. Add images using the second and third bullet points' format from above
9. Save the block
10. Confirm that the newly added images are visible
11. Reopen the block
12. Confirm that all the links appear as static links and their names appear correctly
13. Save block
14. Confirm that all images are still visible

Repeat the steps in the text editor and advanced problem editor. Also, test opening and save existing content for regular problem editor.

PR #501, opened 2024-08-06
fix: asset name parsing in static converter

KristinAoki:
## Description

This PR fixes images not loading in the visual text and problem editor. When assets were loaded with `asset-v1:org+number+run+type@asset+block/image_name.png` asset format, it would incorrectly parse to `asset-v1:org+number+run+type@asset+block/asset-v1:org+number+run+type@asset+block@image_name.png`. A similar issue was fixed in #497. The same src parser is now used for `replaceStaticWithAsset` as was added for `setAssetToStaticUrl`.

## Testing

1. Open an html block
2. Add an image
3. Click the "HTML" button in the toolbar
4. Replace the second "@" with "/"
5. Click "Save"
6. Confirm that the image is still visible


rayzhou-bit (approved):
lgtm

PR #507, opened 2024-09-03
feat!: this repo is unsupported

kdmccormick:
some docs updates, but with a feat! commit
so that it triggers a release and pushes
our "Moved" message to NPM


bradenmacdonald:
Thanks @kdmccormick !

